PR #118, opened 2021-01-22
Remove Hyperledger Fabric 2.0 tests and docs

dzikowski:


docs/schema.json
```diff
@@ -17,10 +17,6 @@
       "title": "Network settings",
       "description": "Basic settings of the Hyperledger Fabric network",
       "type": "object",
-      "default": {
-        "tls": false,
-        "fabricVersion": "1.4.4"
-      },
```
Hejwo, line 23: hm... why default was removed?

docs/schema.json
```diff
@@ -40,12 +36,7 @@
             "1.4.5",
             "1.4.6",
             "1.4.7",
-            "1.4.8",
-            "2.0.1",
-            "2.1.0",
-            "2.1.1",
-            "2.2.0",
-            "2.2.1"
+            "1.4.8"
```
Hejwo, line 39: :+1:

docs/schema.json
dzikowski: because it is already specified for fields in lines 50/41 and 56/47

Hejwo (approved):

PR #127, opened 2021-01-25
Add loglevel to fabrica config

dzikowski:


docs/schema.json
```diff
@@ -45,6 +45,17 @@
           "title": "Use TLS",
           "type": "boolean",
           "default": true
+        },
+        "loglevel": {
```
Hejwo, line 49: I've got mixed feelings about this PR...  
Feature seems logical and might be helpful, but on the other hand setting up different log level is just a matter of changing one line in `.env` file.  

Hyperledger is really complexed and has got tons of options. ei. we could also think about
extraction Orderers MaxMessageCount and wait times, polices, detailed CA settings.  
BUT every change would make config file bigger and bigger. IMO config file should be as concise as possible and contain only most crucial informations. Otherwise we would end up with moving whole complexity of Fabric into config file.

Hejwo: How we could handle such small helper-features:  
- Idea 1: Introduce pre/post generation hooks and leave it to users.
- Idea 2: Most powerful option (but epic task also) would be possibility of extending/overriding our templates or generators  
- Idea 3: Reasonable set of defaults to minimalize fabrica-config file.

samples/fabricaConfig-1org-1channel-1chaincode.json
```diff
@@ -2,7 +2,8 @@
   "$schema": "https://github.com/softwaremill/fabrica/releases/download/0.0.1/schema.json",
   "networkSettings": {
     "fabricVersion": "1.4.6",
-    "tls": false
+    "tls": false,
+    "loglevel": "debug"
```
Hejwo, line 6: czy chcemy żeby to było w debug?

Hejwo (approved):

PR #128, opened 2021-01-25
Verify AND endorsement policy for two orgs

dzikowski:


dzikowski:
already checked in `test-03-private-data.sh`: https://github.com/softwaremill/fabrica/blob/main/e2e-network/test-03-private-data.sh

PR #157, opened 2021-03-08
Init network with simple chaincode

dzikowski:


Dockerfile
```diff
@@ -3,10 +3,15 @@ FROM node:12.18.0-alpine3.12
...
 COPY generators /fabrica/generators
 COPY package.json /fabrica/package.json
 COPY package-lock.json /fabrica/package-lock.json
 
+# copy files for init network
+COPY samples/fabricaConfig-1org-1channel-1chaincode.json /fabrica/generators/init/templates/fabrica-config.json
+COPY samples/chaincodes/chaincode-kv-node /fabrica/generators/init/templates/chaincodes/chaincode-kv-node
+
 WORKDIR /fabrica
 RUN npm install --silent
```
Hejwo, line 16: Probably not for this PR, but I've just noticed it - we copy `generators /fabrica/generators` invoke `npm install` quite early. And they are frequently changed.  
If we postpone it to later we could save some image build time thanks to docker cache.

e2e-network/test-01-simple.sh
```diff
@@ -4,12 +4,9 @@ TEST_TMP="$(rm -rf "$0.tmpdir" && mkdir -p "$0.tmpdir" && (cd "$0.tmpdir" && pwd
 TEST_LOGS="$(mkdir -p "$0.logs" && (cd "$0.logs" && pwd))"
 FABRICA_HOME="$TEST_TMP/../.."
 
-# testing relative path
-CONFIG="../../samples/fabricaConfig-1org-1channel-1chaincode.json"
-
 networkUpAsync() {
   "$FABRICA_HOME/fabrica-build.sh" &&
-    (cd "$TEST_TMP" && "$FABRICA_HOME/fabrica.sh" generate "$CONFIG") &&
+    (cd "$TEST_TMP" && "$FABRICA_HOME/fabrica.sh" init) &&
```
Hejwo, line 9: :+1:

generators/init/index.js
```diff
@@ -5,21 +5,15 @@
...
-    );
+    this.fs.copy(this.templatePath(), this.destinationPath());
 
     this.on('end', () => {
       this.log('===========================================================');
       this.log(chalk.bold('Sample config file created! :)'));
-      this.log('');
-      this.log(`Chaincode directory is '${simpleFabricaConfigJson.chaincodes[0].directory}'.`);
-      this.log('If it\'s empty your network won\'t run entirely.');
+      this.log('You can start your network with \'fabrica up\' command');
```
Hejwo, line 16: Well, overall propositions looks, but why using JS chaincode is an only option?  
IMO it's quite unfair for java chaincode as it is a valid and working option and it might be users choice (same with GO not added yet, but equally important).  

My idea/recommendation:  
- `fabrica init` : initializes fabrica-config.json and copies node chaincode (default option)  
- `fabrica init -l java`: does the same, but for java chaincode

Would suggest adding it in this PR or the next one (but really soon).

e2e/fabricaCommands.test.js
```diff
@@ -49,7 +52,7 @@ describe('validate', () => {
     expect(commandResult).toEqual(TestCommands.success());
     expect(commandResult.output).toContain('Validation errors count: 0');
     expect(commandResult.output).toContain('Validation warnings count: 0');
-    expect(commands.getFiles()).toEqual(['./e2e/__tmp__/commands-tests/fabrica-config.json']);
+    expect(commands.getFiles()).toContain('./e2e/__tmp__/commands-tests/fabrica-config.json');
```
Hejwo, line 55: hm... didn't quite get the case where:
1. user first creates own chaincode in `chaincodes/chaincode-kv-node`
2. then executes `fabrica init`
  
What would happen then? Files will get overridden or we revive yeoman prompt?  

(If it's not handled) Maybe he should get suitable prompt that chaincode already exists and it's advised to manually check if `chaincode.lang` is correct in fabrica-config ?

Hejwo (changes requested):

e2e/fabricaCommands.test.js
dzikowski: to be tested

Dockerfile
dzikowski: we decided not to change it

generators/init/index.js
dzikowski: will be resolved in #158

Hejwo (approved):

PR #177, opened 2021-04-22
Support fabrica config in YAML

dzikowski:


README.md
```diff
@@ -68,11 +68,11 @@ Option `node` makes Fabrica to generate a sample Node.js chaincode as well.
 ### generate
 
 ```bash
-fabrica generate [/path/to/fabrica-config.json [/path/to/fabrica/target]]
+fabrica generate [/path/to/fabrica-config.json|yaml [/path/to/fabrica/target]]
```
Hejwo, line 71: lgtm, but two things before we merge it:
- we should update changelog
- we can improve `fabrica init` now and maybe add a flag `fabrica init --yml`. This can be done in separate issue, but we should remoember about it :)

Hejwo (approved):

PR #184, opened 2021-05-08
Issue 183: Use appropriate Node and chaincode lib versions for Fabric 1.4 and 2.x

dzikowski:


dzikowski:
To consider: select node version depending on fabric version

Hejwo (approved):

PR #222, opened 2021-10-02
Fablo REST support

dzikowski:


e2e-network/expect-invoke-tls.sh
```diff
@@ -1,49 +0,0 @@
-#!/usr/bin/env bash
```
Hejwo, line 1: I'm not a huge fan of this. Using expect-invokes on Fablo Rest is great and simplifies a lot of things, but imo we should still have some raw cli calls with and without tls in tests.
Otherwise we might break something. Invoking/Quering Fabric using cli is different beast and ei. requires proper cert setup in env variables and mounted files

e2e/__snapshots__/fablo-config-hlf1.4-2orgs-1chaincode.json.test.ts.snap
```diff
@@ -920,6 +933,10 @@ services:
       - ../fabric-config/crypto-config/peerOrganizations/org2.com:/var/hyperledger/cli/crypto/
       - ../fabric-config/chaincode-packages:/var/hyperledger/cli/chaincode-packages/
       - \\"$CHAINCODES_BASE_DIR/./chaincodes/chaincode-kv-node-1.4/:/var/hyperledger/cli/chaincode1/\\"
+    depends_on:
+      - ca.org2.com
+      - peer0.org2.com
+      - peer1.org2.com
```
Hejwo, line 939: do we really need depends_on and list of so many machines?
Fabric is distributed system and eventually, at some point, everything should run correctly and get in sync. So ca, orderer, peers don't need to be tightly coupled and don't need each other so much.

e2e/__snapshots__/fablo-config-hlf2-2orgs-2chaincodes-raft.yaml.test.ts.snap
```diff
@@ -947,6 +954,36 @@ services:
...
+      - ca.org1.com
+      - peer0.org1.com
+      - peer1.org1.com
+    networks:
+      - basic
+
+  fablo-rest.org1.com:
+    container_name: fablo-rest.org1.com
+    image: softwaremill/fablo-rest:\${FABLO_REST_VERSION}
+    tty: true
```
Hejwo, line 967: why o we need `tty` ?

e2e/fabloCommands.test.ts
```diff
@@ -32,6 +32,22 @@ describe("init", () => {
     ]);
     expect(commands.getFileContent("fablo-config.json")).toMatchSnapshot();
   });
+
+  it("should init simple fablo config with node chaincode and rest api", () => {
+    // When
+    const commandResult = commands.fabloExec("init node rest");
```
Hejwo, line 38: :+1:

src/extend-config/index.ts
```diff
@@ -15,13 +15,13 @@ const extendConfig = (json: FabloConfigJson): FabloConfigExtended => {
     chaincodes: chaincodesJson,
   } = json;
 
-  const capabilities = configTransformers.getNetworkCapabilities(networkSettingsJson.fabricVersion);
+  const networkSettings = configTransformers.transformNetworkSettings(networkSettingsJson);
+  const capabilities = configTransformers.getNetworkCapabilities(networkSettings.fabricVersion);
   const rootOrg = configTransformers.transformRootOrgConfig(rootOrgJson);
-  const orgs = configTransformers.transformOrgConfigs(orgsJson);
+  const orgs = configTransformers.transformOrgConfigs(orgsJson, networkSettings);
```
Hejwo, line 21: hm... why org needs whole networkSettings block? Can't we just pass `fabloRest` ?

e2e/__snapshots__/fablo-config-hlf2-2orgs-2chaincodes-raft.yaml.test.ts.snap
dzikowski: removed, fablo rest does not need it.

cli, however, does...

PR #252, opened 2021-12-14
Support hlf explorer

gzhk:
Review and merge after https://github.com/softwaremill/fablo/pull/251

### DIFF
--- a/src/types/ConnectionProfile.ts
+++ b/src/types/ConnectionProfile.ts
@@ -1,12 +1,15 @@
-import { NetworkSettings, OrgConfig, PeerConfig } from "./FabloConfigExtended";
+import { ChannelConfig, NetworkSettings, OrgConfig, PeerConfig } from "./FabloConfigExtended";
 
-interface ConnectionProfile {
+interface BaseConnectionProfile {
   name: string;
   description: string;
   version: string;
+  peers: { [key: string]: Peer };
+}
+
+interface ConnectionProfile extends BaseConnectionProfile {
   client: Client;
   organizations: { [key: string]: Organization };
-  peers: { [key: string]: Peer };
   certificateAuthorities: { [key: string]: CertificateAuthority };
 }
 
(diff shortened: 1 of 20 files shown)


src/types/ConnectionProfile.ts
```diff
@@ -120,3 +189,49 @@ export function createConnectionProfile(
...
+    },
+    organizations: {
+      [org.name]: {
+        mspid: org.mspName,
+        adminPrivateKey: {
+          path: `${rootPath}/peerOrganizations/${org.domain}/users/Admin@${org.domain}/msp/keystore/priv-key.pem`,
+        },
+        peers: Object.keys(peers),
+        signedCert: {
+          path: `${rootPath}/peerOrganizations/${org.domain}/users/Admin@${org.domain}/msp/signcerts/Admin@${org.domain}-cert.pem`,
```
Hejwo, line 230: hm... been thinking - we're using here absolute path, don't we?  
Maybe it's possible ti switch to relative?

gzhk: Yes, it's an absolute path. This one is for the explorer, so I guess we could do that, but why should we change it? Do you mean just here, in `createHyperledgerExplorerConnectionProfile` method, or in other methods form this file as well?

Hejwo: I meant only in this place :).  
and why relative is in imo better: someone might want to do `fablo generate` and only use our generated files as a starting point and then customize them and include in their git repository.  
It that scenario he will have absolute paths in commited files.

Hejwo: But bare in mind that it's not super important to fix all places. We had some problems with relative path to chaincodes and thats why they are absolute. It's just an attempt to limit such cases if possible.

gzhk: This path is used by the explorer, inside docker container, so if I get it right, here instead of `/tmp/crypto` we would end up with something close to `../../../../../../../` 😉   imo that would require adjusting as well, so I wouldn't worry about that.

Hejwo (approved):

PR #260, opened 2021-12-17
Update test-on-push.yml

ryjones:
Add workflow_dispatch

Signed-off-by: Ry Jones <ry@linux.com>

### DIFF
--- a/.github/workflows/test-on-push.yml
+++ b/.github/workflows/test-on-push.yml
@@ -1,5 +1,11 @@
 name: Tests
-on: [ push ]
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+  workflow_dispatch:
+
 jobs:
   test-main:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/test-on-push.yml
```diff
@@ -1,5 +1,11 @@
 name: Tests
-on: [ push ]
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+  workflow_dispatch:
+
```
Hejwo, line 8: It looks good :+1:   
Could you please by the way remove from the same file whole `test-dco-sign-off` section with it's sub items?  
We validate DCO on PR creation level anyways and other ways seem to be unnecessary and buggy (like in case of this PR)

Hejwo (approved):

.github/workflows/test-on-push.yml
ryjones: Done!

PR #358, opened 2022-11-05
Fablo k8 fixes and CI

umegbewe:
Signed-off-by: Umegbewe Nwebedu <nwebedujunior55@gmail.com>


umegbewe:
Much deletion due to removing unneeded files forked into the k8 setup

e2e-network/k8s/test-01-simple-k8s.sh
```diff
@@ -20,11 +20,7 @@ dumpLogs() {
 
 networkDown() {
   rm -rf "$TEST_LOGS"
-  (cd "$TEST_TMP" && "$FABLO_HOME/fablo-target/fabric-k8s.sh" down)
-}
-
-chainCode() {
-  (cd "$TEST_TMP" && "$FABLO_HOME/fablo-target/fabric-k8s.sh" chaincodes)
+  (cd "$TEST_TMP" && "$(find . -type f -iname 'fabric-k8s.sh')" down)
```
dzikowski, line 23: why this way? I think it would be better to use file path

e2e-network/k8s/test-01-simple-k8s.sh
```diff
@@ -60,22 +56,23 @@ waitForContainer "$peer0" "Membership view has changed. peers went online:"
 waitForContainer "$peer1" "Learning about the configured anchor peers of Org1MSP for channel my-channel1"
 waitForContainer "$peer1" "Membership view has changed. peers went online:"
 
+
 #Test simple chaincode
-expectInvoke "admin" "org1-peer1.default" "chaincode1" "my-channel1" \
-  "put -a '[\"name\"]' -a Willy Wonka" "{\"success\":\"OK\"}"
-expectInvoke "admin" "org1-peer1.default" "chaincode1" "my-channel1" \
-  "get -a '[\"name\"]'" '{"success":"Willy"}'
+expectInvoke "admin" "org1-peer1.default" "my-channel1" "chaincode1" \
```
dzikowski, line 61: peer domain names should come from fablo-config file - does it work this way? Do we have `org1-peer1.default` in fablo-config file for this test?

e2e-network/k8s/wait-for-chaincode.sh
```diff
@@ -5,10 +5,10 @@ peer=$2
 channel=$3
 chaincode=$4
 version=$5
-config=/home/great/work/new/fablo/fablo-target/fabric-config/fabric-k8/org1.yaml
```
dzikowski, line 8: I have mixed feelings about that - are we able to provide a path than can be used both locally and in CI?

src/setup-k8s/templates/fabric-k8s/scripts/base-functions.sh
```diff
@@ -165,18 +165,25 @@ checkDependencies() {
   if [[ $(command -v kubectl) ]]; then
     printf "\nKubectl installed...\n"
   else
-    printf "\nCouldn't detect a Kubectl \n" && exit
+    printf "\nCouldn't detect Kubectl \n" && exit 1
   fi
 
-  if [[ $(command -v kubectl-hlf) ]]; then
+  if [[ $(export PATH="${KREW_ROOT:-$HOME/.krew}/bin:$PATH" && command -v kubectl-hlf) ]]; then
```
dzikowski, line 171: I believe we should not update `PATH` by fablo. If this is required in CI, then it should be done in CI yaml when we install k8.

e2e-network/k8s/wait-for-chaincode.sh
umegbewe: Wasn't able to find a path that works on both environment,  I think passing the get config path operation to  ``find`` leaves it accurate in most cases.

e2e-network/k8s/test-01-simple-k8s.sh
umegbewe: Took a look at the fablo-config file, peer domain names are not found. What i see:

```   
"peer": {
        "instances": 2,
        "db": "LevelDb"
},
```

In org1.yaml
```
peers:
  org1-peer0.default:
    grpcOptions:
      allow-insecure: false
      hostnameOverride: ""
      ssl-target-name-override: ""
    tlsCACerts:
      pem: |
    url: grpcs://172.18.0.2:31734
  org1-peer1.default:
    grpcOptions:
      allow-insecure: false
      hostnameOverride: ""
      ssl-target-name-override: ""
    tlsCACerts:
      pem: |
    url: grpcs://172.18.0.2:31824
```

src/setup-k8s/templates/fabric-k8s/scripts/base-functions.sh
umegbewe: fixed in latest commit

e2e-network/k8s/test-01-simple-k8s.sh
dzikowski: have a look how domain names are created for docker configuration: https://github.com/umegbewe/fablo/blob/k8/src/setup-docker/templates/fabric-docker/docker-compose.yaml

we have for instance `<%= peer.address %>` or `<%= peer.fullAddress %>` available. If it is possible, it should do it the same way for kubernetes (we should take values from `<%= peer.address %>` in our scripts)

e2e-network/k8s/wait-for-chaincode.sh
dzikowski: I believe `(cd "$TEST_TMP" && "$FABLO_HOME/fablo.sh" down)` should work. And we should support it this way

e2e-network/k8s/test-01-simple-k8s.sh
dzikowski: I believe (cd "$TEST_TMP" && "$FABLO_HOME/fablo.sh" down) should work. And we should support it this way

e2e-network/k8s/wait-for-chaincode.sh
dzikowski: But to be honest, a better way would be to specify config path as a parameter. It would make the script more versatile

e2e-network/k8s/test-01-simple-k8s.sh
umegbewe: I'm quite interested how we can do this in a bash script, to use this template values i think it should be at time of generation,  not sure we are generating bash scripts.

for e.g
expectInvoke "admin" "<%= peer.address %>" "my-channel1" "chaincode1"?

PR #380, opened 2023-03-01
typo fix in readme.md

caner-emec:
Before:

Screenshot from 2023-03-01 13-22-04

After change:

Screenshot from 2023-03-01 13-21-37


dzikowski (approved):
thanks!

PR #403, opened 2023-08-13
This is the first solution to implementing support for invoke command 

jnjerin:
This PR fixes a few issues caused by npm and implements support for invoke command for the simplest use case: 

1. Fixes errors caused by npm:
2. Created a new script `chaincode-script.sh ` to handle `docker exec` 
3. Created the invoke command for its simplest use case


e2e-network/docker/expect-invoke-cli.sh
```diff
@@ -1,7 +1,9 @@
 #!/usr/bin/env bash
 
+echo "FABLO_HOME: $FABLO_HOME"
+
```
dzikowski, line 4: I think those two lines can be removed

src/setup-docker/templates/fabric-docker/chaincode-scripts.sh
```diff
@@ -0,0 +1,53 @@
...
+    $peerAddresses \
+    -C "$channel" \
+    -n "$chaincode" \
+    -c "$command" \
+    --transient "$transient" \
+    --waitForEvent \
+    --waitForEventTimeout 90s \
+    2>&1 
+
+  echo "Executing chaincode invoke command..."
```
dzikowski, line 52: Redundant log message

src/setup-docker/templates/fabric-docker/chaincode-scripts.sh
```diff
@@ -0,0 +1,53 @@
...
+  local peers="$3"
+  local command="$4"
+  local transient="$5"
+
+  echo "-> Chaincode invoke:"
+  echo "Channel: $channel"
+  echo "Chaincode: $chaincode"
+  echo "Peers: $peers"
+  echo "Command: $command"
+  echo "Transient: $transient"
```
dzikowski, line 16: I think we should use a consistent convention, like the one we use for instance [here](https://github.com/hyperledger-labs/fablo/blob/main/src/setup-docker/templates/fabric-docker/scripts/chaincode-functions-v2.sh):

```
chaincodePackage() {
  local CLI_NAME=$1
  local PEER_ADDRESS=$2
  local CHAINCODE_NAME=$3
  local CHAINCODE_VERSION=$4
  local CHAINCODE_LABEL="${CHAINCODE_NAME}_$CHAINCODE_VERSION"
  local CHAINCODE_LANG=$5

  echo "Packaging chaincode $CHAINCODE_NAME..."
  inputLog "CHAINCODE_VERSION: $CHAINCODE_VERSION"
  inputLog "CHAINCODE_LANG: $CHAINCODE_LANG"
  inputLog "PEER_ADDRESS: $PEER_ADDRESS"
  inputLog "CLI_NAME: $CLI_NAME"
```

So:
* snake case with capital letters for variable names
* and `inputLog` instead of echo

dzikowski (approved):

PR #445, opened 2024-05-07
Install chaincode from docker image

sroopsai:


src/setup-docker/templates/fabric-docker/scripts/chaincode-functions-v2.sh
```diff
@@ -12,64 +12,69 @@ dockerPullIfMissing() {
 chaincodeBuild() {
```
Hejwo, line 12: I would recommend to split it into two functions:
`chaincodeBuildDirPath`, `chaincodeBuildImage` It would simplify the logic chaincode-install-v2

src/setup-docker/templates/fabric-docker/commands-generated/chaincode-install-v2.sh
```diff
@@ -8,11 +8,19 @@
...
   "<%= chaincode.lang %>" <% -%>
   "$CHAINCODES_BASE_DIR/<%= chaincode.directory %>" <% -%>
   "<%= global.fabricRecommendedNodeVersion %>"
+<% } else if (chaincode.image) { -%>
+chaincodeBuild <% -%>
+  "<%= chaincode.name %>" <% -%>
+  "<%= chaincode.lang %>" <% -%>
+  "<%= chaincode.image %>"
+  "<%= global.fabricRecommendedNodeVersion %>"
+<% } -%>
```
Hejwo, line 23: As I've written below, having two independent functions `chaincodeBuildDirPath`, `chaincodeBuildImage` would simplify this logic, thanks to it we don't need to have if statements inside functions.

Hejwo:
Reminder for later: Remember about adding snapshots when you'll finish the task.

src/setup-docker/templates/fabric-docker/commands-generated/chaincode-install-v2.sh
dzikowski: Also, to be considered - maybe it is even possible to do build->package->install if we have directory, and just a single step if we have docker image. The flow might be slightly different

src/setup-docker/templates/fabric-docker/scripts/chaincode-functions-v2.sh
sroopsai: Ok 😊

src/setup-docker/templates/fabric-docker/commands-generated/chaincode-install-v2.sh
sroopsai: The flow for deploying Chaincode-as-a-Service is build/pull a Docker Image -> Package A Chaincode (only contains connection details) -> Install -> Run the Chaincode-as-a-service Containers.

PR #448, opened 2024-05-21
depricate test-03-private-data test

### LINKED ISSUE #396: Remove private data v1.4 check from CI
Our current CI uses Fabric 1.4.11 to verify private data for chaincodes. However, this test fails, because it requires obsolete version of Node (8), obsolete platform-dependent node packages, that needs to be compiled with obsolete tools. Also, last 1.4.x Fabric version has been published [over 2 years ago](https://github.com/hyperledger/fabric/releases/tag/v1.4.12).

There is no need to perform extensive tests for Fabric 1.4.x. And we should:
- [ ] Remove `test-03-private-data` check - both from CI and the whole test script
- [ ] Move private data checks to `test-02-private-data`
- [ ] Rename remaining checks, to keep naming consistent (i.e. either rename 04 to 03, or remove numbers entirely from names)
- [ ] Update `TEST_CASES.md` to reflect the updated test cases

Sanket-0510:
fixes #396 
I have removed the test-03-private-data from the CI and have also renamed couple of files in CI as well as in script.
- [x] Remove test-03-private-data check - both from CI and the whole test script
- [x] Move private data checks to test-02-private-data
- [x] Rename remaining checks, to keep naming consistent (i.e. either rename 04 to 03, or remove numbers entirely from names)
- [x] Update TEST_CASES.md to reflect the updated test cases


Sanket-0510:
@dzikowski @Hejwo one doubt I have is that do  we still needs private-data checks in the CI but under some other job?

dzikowski:
Yes, preferrably in test-02

Sanket-0510:
> Yes, preferrably in test-02

you mean in test-02-Raft job? isn't that specially for raft?

dzikowski:
yes, the raft one. Those tests are too heavy to have a separated one for private data, so I think it is fine to combine it with raft. Also have a look at the checklist here: https://github.com/hyperledger-labs/fablo/issues/396

dzikowski:
resolved in https://github.com/hyperledger-labs/fablo/pull/467

Issue #346, opened 2022-08-16
Installing the same chaincode on two channels


rrozek:
Hi, tremendous progress in fablo 1.1.0!
However i noticed, that I cannot install the same chaincode on two separate channels.
My config is for 2 orgs, 2 peer per org, 1 channel per org and i want to install the same chaincode on both channels. It fails on:
```
Generating network config
    FABLO_VERSION:      1.1.0
    FABLO_CONFIG:       custom-config-2org-2peer-1channel-per-org.json
    FABLO_NETWORK_ROOT: /home/jrojek/dvnt/repo/open-pay-api/configs/fablo/fablo-target
Validation errors count: 1
Validation warnings count: 0
===========================================================
=================== Validation summary ====================
Errors found:
  Chaincode:
   - Chaincode name 'chaincode-test' is not unique.
===========================================================
```

config i tried:
```
{
  "$schema": "https://github.com/hyperledger-labs/fablo/releases/download/1.1.0/schema.json",
  "global": {
    "fabricVersion": "2.4.3",
    "tls": true
  },
  "orgs": [
    {
      "organization": {
        "name": "Orderer",
        "mspName": "OrdererMSP",
        "domain": "root.com"
      },
      "ca":  {
        "prefix": "ca"
      },
      "orderers": [
        {
          "groupName": "group1",
          "type": "raft",
          "instances": 3
        }
      ],
    },
    {
      "organization": {
        "name": "Org1",
        "mspName": "Org1MSP",
        "domain": "org1.com"
      },
      "ca":  {
        "prefix": "ca"
      },
      "peer": {
        "instances": 2,
        "anchorPeerInstances": 2,
        "db": "CouchDb"
      },
      "tools": {
        "fabloRest": true
      }
    },
    {
      "organization": {
        "name": "Org2",
        "mspName": "Org2MSP",
        "domain": "org2.com"
      },
      "ca":  {
        "prefix": "ca"
      },
      "peer": {
        "instances": 2,
        "anchorPeerInstances": 2,
        "db": "CouchDb"
      },
      "tools": {
        "fabloRest": true
      }
    }
  ],
  "channels": [
    {
      "name": "channel1-org1",
      "orgs": [
        {
          "name": "Org1",
          "peers": [
            "peer0",
            "peer1"
          ]
        }
      ]
    },
    {
      "name": "channel1-org2",
      "orgs": [
        {
          "name": "Org2",
          "peers": [
            "peer0",
            "peer1"
          ]
[...]

Issue #228, opened 2021-10-18
Fablo fails if two chaincodes have the same name


dzikowski:
To reproduce: create a config with two chaincodes with the same name (on different channels).

The network won't start, since we do not aim (yet?) to support chaincodes with the same names (i fails with duplicate volumes).

AND when you try to prune the network, the prune operation fails as well (duplicate volumes as well).

dzikowski:
Partially resolved - user gets the proper validation information in case of duplicate chaincode names (#596)

PR #457, opened 2024-06-22
Fix: Support Channel-Specific Chaincode Directories to Prevent Duplicate Docker Mounts

SkySingh04:
Fix: updated fablo.sh to support channel-name/chaincode-name directory structure for Docker mounts

- Modify `executeOnFabloDocker` function to dynamically create Docker volume mounts for each chaincode based on the channel-name/chaincode-name directory structure.
- This change resolves the issue of duplicate Docker mount errors when multiple chaincodes are defined for the same network.

Fixes: #346 #228

### DIFF
--- a/fablo.sh
+++ b/fablo.sh
@@ -135,11 +135,21 @@ executeOnFabloDocker() {
 
     fablo_config="$(cd "$(dirname "$fablo_config")" && pwd)/$(basename "$fablo_config")"
     local chaincodes_base_dir="$(dirname "$fablo_config")"
+
+    # Extract chaincodes information from the config file
+    local chaincodes=$(jq -r '.chaincodes[] | "\(.channel)/\(.name)"' "$fablo_config")
+
     fablo_config_params=(
       -v "$fablo_config":/network/fablo-config.json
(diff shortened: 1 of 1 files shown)


Akash-Singh04:
@dzikowski Kindly review the PR

dzikowski:
The proposed solution does not solve the issue. The actual error is not about mounting additional volumes for Fablo container. It is about mounting names in Docker Compose file for Fabric CLI container. Have a look at this Docker Compose template: https://github.com/hyperledger-labs/fablo/blob/main/src/setup-docker/templates/fabric-docker/docker-compose.yaml#L102

```
      <%_ chaincodes.forEach(function(chaincode) { _%>
      - "$CHAINCODES_BASE_DIR/<%= chaincode.directory %>/:/var/hyperledger/cli/<%= chaincode.name %>/"
```

Instead of target directory pattern `/var/hyperledger/cli/<chaincode-name>`, we want to have `/var/hyperledger/cli/<channel-name>-<chaincode-name>`, or `/var/hyperledger/cli/<channel-name>/<chaincode-name>`.

Then we need to ensure everything works - probably enough is running locally `e2e-network/docker` scripts.

Issue #507, opened 2025-02-07
Fix documentation: chaincode invoke


SRshohan:
To invoke chaincode:

`
fablo chaincode invoke <channel-name> <chaincode-name> <peers-domains-comma-separated>  <command> [transient]
`

swap 
`fablo chaincode invoke <peers-domains-comma-separated>  <channel-name>  <chaincode-name> <command> [transient]
`

PR #508, opened 2025-02-07
Swapped channelName with peer-domain

SRshohan:
Fix documentation: chaincode invoke #507

The documentation incorrectly used `channelName` where `peer-domain` should be used when invoking chaincode.  This commit swaps the incorrect `channelName` with the correct `peer-domain` value.

### DIFF
--- a/README.md
+++ b/README.md
@@ -213,12 +213,12 @@ Chaincode directory is specified in Fablo config file.
 Invokes chaincode with specified parameters.
 
 ```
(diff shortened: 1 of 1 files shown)


dzikowski:
thanks for the fix @SRshohan! could you change the commit to include DCO sign-off, and then force push? We have a requirement that all commits need to include it. https://wiki.linuxfoundation.org/dco

SRshohan (commented):
I added DCO sign-off

dzikowski (approved):

PR #521, opened 2025-04-12
Fix #519: Check if post-generate.sh exists before executing

### LINKED ISSUE #519: Misleading error on starting invalid network
When Fablo starts a network, and there is an error in config file (like raft with no tls), the `/hooks/post-generate.sh` file is not created. Which leads to a misleading error on this line in `fablo.sh`:

```sh
("$fablo_target/hooks/post-generate.sh")
```

Output:

```
=================== Validation summary ====================
Errors found:
  Orderer:
   - Raft consensus type must use network in TLS mode. Try setting 'global.tls' to true
Warnings found:
  Orderer:
   - Orderer consesus type is set to 'raft', but number of instances is 1. Network won't be fault tolerant! Consider higher value.
===========================================================
fablo:setup-network
Formatting generated files
../fablo.sh: line 195: /Users/xxx/fablo/samples/fablo-target/hooks/post-generate.sh: No such file or directory
```

So, there should be a check first in`fablo.sh` if the file exists.

VijetaPriya47:
This PR fixes issue #519 where Fablo shows a misleading error when starting an invalid network.

Issue:- The problem was that when there's an error in the config file (like RAFT with no TLS), the post-generate.sh file is not created, but the script tries to execute it anyway, resulting in a misleading error message.

Changes made:
- Added a check to verify if post-generate.sh exists before executing it
- The script now silently skips the post-generate hook if it doesn't exist

Fixes #519
I would be happy to make any further changes in the code.
Screenshot from 2025-04-12 12-05-10

### DIFF
--- a/fablo.sh
+++ b/fablo.sh
@@ -191,7 +191,11 @@ generateNetworkConfig() {
 
   mkdir -p "$fablo_target"
   executeOnFabloDocker "fablo:setup-network" "$fablo_target" "$fablo_config"
-  ("$fablo_target/hooks/post-generate.sh")
+  if [ -f "$fablo_target/hooks/post-generate.sh" ]; then
+    ("$fablo_target/hooks/post-generate.sh")
+  else
+    echo "Note: No post-generate.sh hook found. This might be due to validation errors in your configuration."
+  fi
 }
 
 networkPrune() {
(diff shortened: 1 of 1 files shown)


VijetaPriya47:
Hi @dzikowski,

I apologize for the oversight — I initially didn’t notice that the application was intended for Intern 2025. However, I’ve corrected the relevant details in the pull request.

I would be grateful if you could kindly review and approve it at your convenience.

dzikowski:
@VijetaPriya47 don't worry about labels. It's just marked as a potential issue for internship. The internship itself is going to be broader, heavily focused on feature parity, and only some of the issues with the label are going to be actually part of the internship.

dzikowski (approved):

PR #523, opened 2025-04-12
Fix #340: Connection profile - bugs and missing values

### LINKED ISSUE #340: Connection profile - bugs and missing values
1. CA URL in connection profile starts from `http://`, even if TLS is enabled (it should be `https://` in this case)
2. missing `orderers` section (required when service discovery is disabled - for instance in dev mode)
3. missing `channels` section (required when service discovery is disabled)

When all above issues are fixed, a Node.js client could use it with service discovery disabled

VijetaPriya47:
This PR fixes issue #340 by addressing the following issues with connection profiles:

1. CA URL in connection profile starts from `http://`, even if TLS is enabled
   - Fixed by ensuring the protocol is set based on TLS settings

2. Missing `orderers` section (required when service discovery is disabled)
   - Added orderers section to connection profiles

3. Missing `channels` section (required when service discovery is disabled)
   - Added channels section to connection profiles

These changes ensure that Node.js clients can use the connection profiles with service discovery disabled, which is particularly important in dev mode.

Fixes #340

### DIFF
--- a/src/setup-docker/index.ts
+++ b/src/setup-docker/index.ts
@@ -54,7 +54,7 @@ export default class SetupDockerGenerator extends Generator {
 
     // ======= fabric-config ============================================================
     this._copyOrgCryptoConfig(orgs);
-    this._createConnectionProfiles(global, orgs);
+    this._createConnectionProfiles(global, orgs, channels, config.ordererGroups);
     this._createFabricCaServerConfigs(orgs);
     this._createExplorerMaterial(global, orgs, channels);
     this._copyConfigTx(config);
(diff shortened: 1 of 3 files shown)


src/setup-docker/index.ts
```diff
@@ -101,9 +101,9 @@ export default class SetupDockerGenerator extends Generator {
     });
   }
 
-  _createConnectionProfiles(global: Global, orgsTransformed: OrgConfig[]): void {
+  _createConnectionProfiles(global: Global, orgsTransformed: OrgConfig[], channels: ChannelConfig[] = [], ordererGroups: OrdererGroup[] = []): void {
```
dzikowski, line 104: please require explicit parameters (less prone to errors)
```suggestion
  _createConnectionProfiles(global: Global, orgsTransformed: OrgConfig[], channels: ChannelConfig[], ordererGroups: OrdererGroup[]): void {
```

PR #524, opened 2025-04-12
Fix #111: Add post-start hook functionality

### LINKED ISSUE #111: Add post-start hooks
We could add hooks exectuted after:
- [ ] network startup/end
- [ ] chaincode pre-install / after-install

VijetaPriya47:
This PR fixes issue #111 by adding post-start hook functionality.

Changes made:
- Added a new `postStart` hook type that is executed after the network is started
- Updated the configuration interfaces and extension logic to handle the new hook
- Created template files for the post-start hook
- Updated the network startup scripts to execute the post-start hook
- Updated documentation to explain the new hook functionality

This allows users to define custom actions to be performed after the network is started, which can be useful for various post-startup tasks like sending notifications, running additional setup scripts, or deploying applications that depend on the network.

Fixes #111

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 ### Features
 * Full support for Fabric 3.0.0 and drop solo consensus
   [#513](https://github.com/hyperledger-labs/fablo/pull/513)
+* Add post-start hook functionality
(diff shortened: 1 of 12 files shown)


VijetaPriya47:
Okay let me check this again.

PR #544, opened 2025-05-02
Adds gateway option to init

ronaldl29:


src/init/index.ts
```diff
@@ -21,6 +21,14 @@ export default class InitGenerator extends Generator {
       fabloConfigJson = { ...fabloConfigJson, chaincodes: [] };
     }
 
+    const shouldInitWithNodeSampleGateway = this.args.length && this.args.find((v) => v === "gateway");
+    if (shouldInitWithNodeSampleGateway) {
+      console.log("Creating sample Node.js gateway");
+      this.fs.copy(this.templatePath("gateway"), this.destinationPath("gateway"));
+    } else {
+      fabloConfigJson = { ...fabloConfigJson, gateway: [] };
+    }
```
dzikowski, line 30: we don't need the `else` clause here

src/init/index.ts
```diff
@@ -21,6 +21,14 @@ export default class InitGenerator extends Generator {
       fabloConfigJson = { ...fabloConfigJson, chaincodes: [] };
     }
 
+    const shouldInitWithNodeSampleGateway = this.args.length && this.args.find((v) => v === "gateway");
+    if (shouldInitWithNodeSampleGateway) {
+      console.log("Creating sample Node.js gateway");
+      this.fs.copy(this.templatePath("gateway"), this.destinationPath("gateway"));
```
dzikowski, line 27: you need to copy the gateway into fablo docker, the same way as here: https://github.com/hyperledger-labs/fablo/blob/main/Dockerfile#L13

ronaldl29:
Thank you for the pointers! @dzikowski I will get this fixed.

src/init/index.ts
ronaldl29: Fixed

src/init/index.ts
ronaldl29: Does this look okay? @dzikowski

ronaldl29:
Perfect! That sounds good to me @dzikowski

Issue #99, opened 2020-12-07
Create sample chaincode in Go and test it in CI


dzikowski:
Any sample could be used like this one for example

https://github.com/hyperledger/fabric-samples/tree/main/asset-transfer-basic/chaincode-go

joseucarvajal:
+1

debayangg:
@dzikowski I would like to work on this issue. I went through all the samples and saw that this [config](https://github.com/hyperledger-labs/fablo/blob/main/samples/fablo-config-hlf2-2orgs-2chaincodes-raft.yaml) uses two chaincodes one in nodejs and another in java which is later tested [here](https://github.com/hyperledger-labs/fablo/blob/main/e2e-network/docker/test-02-raft-2orgs.sh) . Should I follow these and create a sample chaincode in Go and test it and later add it to GH actions.

dzikowski:
yes, it would be perfect this way! You can even remove the node.js chaincode from that config and replace it with the one in Go. Would be leaner, and we test Node.js chaincode also in other tests

kalio007:
@dzikowski if this is cold, i would love to handle this

debayangg:
@kalio007 was working on this, was able to create a simple golang chaincode with a test but I have been facing problems with the test-02-raft-2orgs.sh integrating go chaincode with java

kalio007:
Thanks for the update, do you mind sharing your current changes to the PR so someone can take a look and help out?

debayangg:
The basic tests itself were failing so I didnt create a PR and was investigating but okay then I can create a draft PR for now,

debayangg:
@dzikowski since I am facing some issues and have to investigate them, can I just create a simple chaincode and yaml setup for golang exclusively and test it with a custom shell script that conforms to the style in e2e-network/docker. Also can you provide some context regarding the e2e tests and snapshots.

dzikowski:
Is the relevant e2e-network test script working? If yes, then it might be just a matter of updating snapshots with `npm run test:e2e-update` (please check also our contribution guide)

debayangg:
Okay I will update the e2e tests and check again

debayangg:
I tried investigating further and it seems to be mainly due to mismatch between fabric and the go chaincode versions, as documented in the following issues:
1. https://github.com/hyperledger/fabric/issues/4951
2. https://github.com/hyperledger/fabric/issues/3039

I have tried go mod vendor but it didn't fix this issue.

[...]

PR #545, opened 2025-05-02
Add testing for Golang Chaincode

debayangg:
Addresses the [Issue #99 ](https://github.com/hyperledger-labs/fablo/issues/99).


debayangg:
@kalio007 @dzikowski the e2e tests are failing due to test snapshots not matching which is due to using of golang in the the samples yaml file. It can be fixed by creating a new file or updating the snapshots but the main problem is that test-02-raft was was also failing when I checked it personally.

It seems to a docker problem, on my system I got :
```
Error: chaincode install failed with status: 500 - failed to invoke backing implementation of 'InstallChaincode': could not build chaincode: docker build failed: docker image build failed: docker build failed: Error returned from build: 2 "google.golang.org/protobuf/internal/flags
```
A simple golang chaincode for setting and getting a key with 1 org, 1 orderer and 1 channel was working fine without errors.

debayangg:
This is the result of test-02-raft in my repo, https://github.com/debayangg/fablo/actions/runs/14825376358/job/41617875740.
The error is again the same,
```
Error: chaincode install failed with status: 500 - failed to invoke backing implementation of 'InstallChaincode': could not build chaincode: docker build failed: docker image build failed: docker build failed: Error returned from build: 2 "go: downloading github.com/hyperledger/fabric-contract-api-go v1.2.2
```

dzikowski:
I've read also your comment on failing Java chaincode for the newest version: https://github.com/hyperledger-labs/fablo/issues/99#issuecomment-2850024895

Isn't it just a matter of upgrading Java library version?
* https://github.com/hyperledger/fabric-chaincode-java/blob/main/COMPATIBILITY.md
* https://github.com/hyperledger-labs/fablo/blob/main/samples/chaincodes/chaincode-java-simple/build.gradle#L25

Do you think Java library upgrade should be a separate issue?

debayangg:
No I dont think so, I will update the jdk version and try again. I had an exam today so didn't get time till now. Overall it all seems to be a problem of incompatible versions.

debayangg:
I went ahead and made the changes in the java fabric sdk, tried out different versions and also tried versions used in chaincode samples provided in the fabric samples, it seems that Java chaincode will also need to be updated. It will take some time on my part to read through Java chaincode and refactor it as simple automated refactors are not working.

dzikowski:
Perfect, thanks!

joaormatos:
> Yes, if I read action logs properly, it seems it requires Go in version 1.19 or higher. The chaincode is executed in a docker container, which uses `hyperledger/fabric-ccenv` image: https://hub.docker.com/r/hyperledger/fabric-ccenv. Fablo uses the same version of the image as the version of Hyperledger Fabric, which is in this case `2.4.3`.
> 
> the output of:
> 
> ```
> docker run -it hyperledger/fabric-ccenv:2.4.3 go version
> ```
> 
> is:
> 
> ```
> go version go1.17.5
> ```
> 
> So I think the solution is to update the Fabric version in relevant fablo config file. You may choose for instance the newest `2.5.*`

I suggest doing the opposite: **modify the sample chaincode so that it's compatible with older versions of Go.**
Specifically, ensure that the sample chaincode targets the **oldest** version of Go that's still used by some version of Fabric, among the versions of Fabric that Fablo intends to support. 

A Go module declares its target version in `go.mod` using the `go` directive; the current sample is declaring Go 1.21 as its target Go version, as seen [here](https://github.com/hyperledger-labs/fablo/pull/545/commits/8c69b74d59e245772c4f8273615f3ca63ee3390e#diff-3d79b1615adcc625d24478ca0a88db0485a97c81676dd10ed257111bbcdbe819R3)

You can try updating the target version like this:

`go mod tidy -go=1.17`

Another thing: when packaging Go chaincode, `go mod vendor` should be executed prior to the packaging itself, to ensure the chaincode package includes all of its dependencies.
It's analogous to what happens when you run `npm install` and npm populates `node_modules` with the project dependencies.

References:
https://go.dev/ref/mod#go-mod-file-go
https://go.dev/ref/mod#go-mod-tidy
https://go.dev/ref/mod#vendoring

Note: Because the `go` directive was only advisory until Go 1.21, older versions like 1.17.x will still try and *some times* succeed in building chaincode packages that target newer versions, provided that the code in the module and its dependencies is effectively compatible with Go 1.17. That sounds like quite the potential source of confusion, so it might be helpful to warn users in such situations.

PR #546, opened 2025-05-04
feat: Add cross-platform testing on macOS

### LINKED ISSUE #217: Test different environments in GithubActions (for example OS X)
see: https://github.com/actions/virtual-environments

OsamaRab3:
Fix #217

### DIFF
--- a/.github/workflows/test-on-push.yml
+++ b/.github/workflows/test-on-push.yml
@@ -40,6 +40,38 @@ jobs:
       - name: Lint
         run: npm run lint && ./lint.sh
 
+
+  test-cross-platform:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest]
+    steps:
+      - name: Check out repository code
+        uses: actions/checkout@v2
+
+      - name: Install dependencies (macOS)
+        if: runner.os == 'macOS'
+        run: |
+          brew install shellcheck yamllint
+          npm install -g npm@latest
+
+      - name: Build Fablo
+        run: |
+          shellcheck --version && \
+          yamllint -v && \
+          npm install && \
+          npm run build && \
+          ./fablo-build.sh
+
+      - name: Run unit tests
+        run: npm run test:unit
+
+      - name: Lint
+        run: npm run lint
+      
+        
   #  test-k8:
   #    needs: test-main
   #    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


OsamaRab3:
@dzikowski   
please review my pull request and let me know if there's anything that needs to be adjusted.

dzikowski:
It seems docker setup is problematic for macos. Also each time you push I need to approve the workflow to be run (this is required by github for first time contributors). Could you experiment on a PR that does not point to main fablo repository? I believe in such case actions are executed immediately, without approvals.

dzikowski:
Unfortunately I don't have clear answers:
1. I don't know, sorry, figuring this out is a part of this issue
2. Fablo starts the Fabric network on Docker - so we need to have a docker in runner
3. The same answer as for question 1. I don't know

umegbewe:
@OsamaRab3 colima works on ``macos-13`` which is Intel basd, everything newer (macos-14, macos-latest, arm64 runners) is Apple-silicon and can't work without nested virtualization.

You can go ahead using the macos-13 runner, an example is colima testing on macos [here](https://github.com/ddev/ddev/blob/main/.github/workflows/colima-tests.yml)

Alternatively you can use the [setup-docker-macos-action](https://github.com/douglascamata/setup-docker-macos-action) which simplifies the colima install and start process

OsamaRab3:
@dzikowski just wanted to check if everything looks good to you, or if there's anything you'd like me to adjust. Thanks!

umegbewe:
@OsamaRab3 turns out we could test on macos-15 arm, needs some tweaking to achieve that see an example [ here ](https://github.com/mikekazakov/nimble-commander/actions/runs/14836389701/workflow#L34), please go ahead with this.

Also i don't think another cross platform test is necessary, we could change one of existing ones to run on mac e.g ``test-02-raft``

OsamaRab3:
> @OsamaRab3 turns out we could test on macos-15 arm, needs some tweaking to achieve that see an example [ here ](https://github.com/mikekazakov/nimble-commander/actions/runs/14836389701/workflow#L34), please go ahead with this.
> 
> Also i don't think another cross platform test is necessary, we could change one of existing ones to run on mac e.g `test-02-raft`
---
Thanks! 
I went ahead and applied this in `test-05-v3`, which now runs on macos-15 with the necessary setup (Docker, Colima, patched Lima, etc).
Let me know if you'd prefer switching an existing test like `test-02-raft`

OsamaRab3:
@dzikowski  
@umegbewe  please review the PR

umegbewe (approved):
Approved with thanks!

PR #551, opened 2025-05-10
Improve config.test.ts

Sweetdevil144:
Add test cases for `config.test.ts` for `isFabloVersionSupported` tag and `versionsSupportingRaft`

### DIFF
--- a/src/config.test.ts
+++ b/src/config.test.ts
@@ -1,4 +1,4 @@
-import { fabloVersion, getVersionFromSchemaUrl } from "./config";
+import { fabloVersion, getVersionFromSchemaUrl, isFabloVersionSupported, supportedVersionPrefix, versionsSupportingRaft } from "./config";
 
 it("should get version from schema URL", () => {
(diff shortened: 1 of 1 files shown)


src/config.test.ts
```diff
@@ -10,3 +10,18 @@ it("should get current version in case of missing schema URL", () => {
...
+  expect(isFabloVersionSupported(`${currentMajorMinor}.999`)).toBeTruthy();
+  expect(isFabloVersionSupported("0.0.1")).toBeFalsy();
+});
+
+it("should check versions supporting raft", () => {
+  expect(versionsSupportingRaft("1.4.3")).toBeTruthy();
+  expect(versionsSupportingRaft("1.4.4")).toBeTruthy();
+  expect(versionsSupportingRaft("2.0.0")).toBeTruthy();
+  expect(versionsSupportingRaft("1.4.2")).toBeFalsy();
+  expect(versionsSupportingRaft("1.3.0")).toBeFalsy();
```
dzikowski, line 26: Fablo actually dropped support for versions below `2.0`, and all versions support RAFT, so this test should be reworked, and also our checks in `./config` file

Issue #138, opened 2021-01-28
Export network topology with PlantUML or Mermaid


dzikowski:

OsamaRab3:
I’d like to clarify a few things before starting:

 1. What all would need to be changed in order to support this feature?
 2. Should the export support both **PlantUML** and **Mermaid**, or is it okay to start with just one?
 3. Where should the export be triggered from — a **CLI command**, a **config option**, or a **separate script**?

 Looking forward to your guidance so I can proceed in the right direction. Thanks!

dzikowski:
I have a feeling Mermaid is more common tool, so that would be sufficient. I believe we need a similar command as `validate` or `extend-config`. It should follow the same pattern

OsamaRab3:
I’ve implemented the initial part of the network `topology export` using `Mermaid` . Here's a generated diagram based on the current implementation:

![Image](https://github.com/user-attachments/assets/24396a88-e7bc-4405-b5e3-140f8d1c68a6)

Could you please confirm if this output matches your expectations in terms of structure and visual representation?

Also, I’m not entirely sure what parts of the codebase still need to be updated to fully integrate this feature with Fablo. Are there specific files or modules I should focus on next to make this work end-to-end

dzikowski:
resolved in #565

PR #565, opened 2025-06-21
Export network topology with  Mermaid 

OsamaRab3:
implementation of export network topology  #138

---
The generated diagram: 

image


src/export-network-topology/generateMermaidDiagram.ts
```diff
@@ -0,0 +1,153 @@
...
+    config.chaincodes.forEach((cc) => {
+      if (cc.channel) {
+        // const channelName = typeof cc.channel === "string" ? cc.channel : (cc.channel as any).name;
+        // const channelId = safeId(`Channel_${channelName}`);
+        // const ccId = safeId(`Chaincode_${cc.name}`);
+        // Optionally, you can skip this connection since chaincode is inside the channel subgraph
+        // diagram += `  ${ccId} -->|on ${channelName}| ${channelId}\n`;
+      }
+    });
+  }
```
dzikowski, line 150: it can be probably removed

src/export-network-topology/index.ts
```diff
@@ -0,0 +1,63 @@
...
+      }
+
+      const configContent = this.fs.read(this.fabloConfigPath);
+      if (!configContent) {
+        throw new Error(`Failed to read configuration file: ${this.fabloConfigPath}`);
+      }
+
+      const json = parseFabloConfig(configContent);
+      const configExtended: FabloConfigExtended = extendConfig(json);
+      const mermaidDiagram = generateMermaidDiagram(configExtended);
```
dzikowski, line 48: for the next PR: I think we should create the diagram for each `generate` command by default. Why - because the operation is cheap in terms of performance, and we get free testing

dzikowski (approved):

src/export-network-topology/index.ts
OsamaRab3: Got it! I’ll include automatic diagram generation in the next PR as suggested

PR #594, opened 2025-08-12
Feat/ccaas

dpereowei:


src/init/index.ts
```diff
@@ -1,13 +1,14 @@
...
+import { version } from "../../package.json";
 
 function getDefaultFabloConfig(): FabloConfigJson {
   return {
-    $schema: "https://github.com/hyperledger-labs/fablo/releases/download/2.2.0/schema.json",
+    $schema: `https://github.com/hyperledger-labs/fablo/releases/download/${version}/schema.json`,
     global: {
       fabricVersion: "2.5.12",
-      tls: false,
+      tls: true,
```
dzikowski, line 11: It is wort adding to the changelog file - that starting from this PR the init command generates network with TLS/RAFT instead of non-TLS/solo

src/setup-docker/templates/fabric-docker/commands-generated.sh
```diff
@@ -28,6 +28,8 @@ generateArtifacts() {
 startNetwork() {
   printHeadline "Starting network" "U1F680"
   (cd "$FABLO_NETWORK_ROOT"/fabric-docker && docker compose up -d)
+  # todo start ccaas containers if needed
+  # todo remore ccaas start from chaincode scripts
```
dzikowski, line 32: redundant todos

src/setup-docker/templates/fabric-docker/commands-generated.sh
```diff
@@ -241,25 +243,26 @@ printStartSuccessInfo() {
 
 stopNetwork() {
   printHeadline "Stopping network" "U1F68F"
+  # todo stop ccaas containers
```
dzikowski, line 246: redundant todo

src/setup-docker/templates/fabric-docker/scripts/base-functions-v3.sh
```diff
@@ -38,6 +38,47 @@ certsGenerate() {
   done
 }
 
+certsGenerateCCaaS() {
```
dzikowski, line 41: it should be probably the same as base-functions-v2

PR #528, opened 2025-04-16
Add check for unique chaincode names across channels

IITI-tushar:
fixes : [issue#228](https://github.com/hyperledger-labs/fablo/issues/228)
This pull request introduces a new validation to ensure that chaincode names are unique across different channels. It includes changes to both the test suite and the core configuration extension logic.

New validation for unique chaincode names:

* [`e2e/fabloCommands.test.ts`](diffhunk://#diff-6e7f9a9fbd6b76432d0144fa0a0524ad512adf3e3e4c90a669c0527fb52bc321R140-R150): Added a test case to check for duplicate chaincode names across different channels, ensuring that the validation throws an error when duplicates are found.
* [`src/extend-config/extendChaincodesConfig.ts`](diffhunk://#diff-410d798b71034d8041b7c7be2f8b7930c9786f2d27dad27085c94ef4a525d97cR40-R55): Introduced the `checkUniqueChaincodeNames` function to validate that all chaincode names are unique within the provided configuration. This function is called within the `extendChaincodesConfig` function.

Integration of the new validation:

* [`src/extend-config/extendConfig.ts`](diffhunk://#diff-824fa2ee75ec65611ab6464fada13e0db18affcf06759816893ee5e61dec5b84R7): Imported the `checkUniqueChaincodeNames` function and integrated it into the `extendConfig` function to ensure chaincode names are checked for uniqueness during the configuration extension process.

PR #596, opened 2025-08-15
Add check for unique chaincode names

OsamaRab3:
This PR continues the work started in PR #528

**_Notes:_**
Original implementation credit goes to the author of PR #528
This PR finalizes and adapts their work .

_TODO:_ 
should close issue #228


dzikowski:
I think we should approach testing in a slightly different way:
1. `samples` directory contains valid samples that people can use. So we should remove the newly created config
2. Instead we should test the config by altering the file on the fly. Test `should throw an error for duplicate chaincode names across different channels` should update the fablo config in a similar way as here: should throw an error for duplicate chaincode names across different channels
3. We can also remove `should validate when all chaincode names are unique across channels`, because there is no added value from it. Validation passes anyway in all cases.
4. Also we can remove `e2e/fablo-config-hlf2-2orgs-2chaincodes-private-data-duplicate.yaml.test.ts` and the newly created snapshot

dzikowski (approved):

Issue #610, opened 2025-09-26
Detect changes in fablo-config.json when network is already generated


dzikowski:
Steps to reproduce:
```
fablo up
fablo down
# change something in fablo-config.json
fablo up
```

Fablo uses old generated files, starting the old network.

Expected behavior:
1. on up fablo should copy fablo config json into fablo-target
2. when fablo up is called second time for different config, it should fail, print diff, and print the message that the user needs to prune the network first

PR #614, opened 2025-09-28
Detect changes in fablo-config.json and prevent reusing old network

OsamaRab3:
fix: #610


fablo.sh
```diff
@@ -28,6 +28,73 @@ getDefaultFabloConfig() {
...
+  cp -f "$fablo_config" "$fablo_target/$preserved_name"
+
+}
+
+getStoredConfigSnapshotPath() {
+  if [ -f "$FABLO_TARGET/fablo-config.json" ]; then
+    echo "$FABLO_TARGET/fablo-config.json"
+  elif [ -f "$FABLO_TARGET/fablo-config.yaml" ]; then
+    echo "$FABLO_TARGET/fablo-config.yaml"
+  else
```
umegbewe, line 52: This will fail on `.yml` extensions, add a condition for that

umegbewe (changes requested):

Issue #604, opened 2025-09-05
Explorer needs restart to start working properly


dzikowski:
Fablo starts all nodes (orderers, peers, and explorer), and after that it performs channel creation etc. This is probably why explorer node needs a restart, after the whole network is bootstrapped to start picking blocks. We need a manual restart of explorer in `commands-generated.sh` once channels are created (can be before chaincode installation).

PR #615, opened 2025-09-28
Restart Explorer

OsamaRab3:
#604

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -86,6 +86,7 @@
       "integrity": "sha512-yDBHV9kQNcr2/sUr9jghVyz9C3Y5G2zUM2H2lo+9mKv4sFgbA8s8Z9t8D1jiTkGoO/NoIfKMyKWr4s6CN23ZwQ==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "@ampproject/remapping": "^2.2.0",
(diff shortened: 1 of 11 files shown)


src/setup-docker/templates/fabric-docker/commands-generated.sh
```diff
@@ -106,6 +106,19 @@ installChannels() {
       <% }) -%>
     <% }) -%>
   <% } -%>
+  
+  # Restart Explorer 
+  printItalics "Restarting Explorer" "U1F504"
```
dzikowski, line 111: I think this line can be removed - we already have logs about restarting global and org explorer, and this particular line gets printed even if explorer is disabled

package-lock.json
```diff
@@ -86,6 +86,7 @@
       "integrity": "sha512-yDBHV9kQNcr2/sUr9jghVyz9C3Y5G2zUM2H2lo+9mKv4sFgbA8s8Z9t8D1jiTkGoO/NoIfKMyKWr4s6CN23ZwQ==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
```
dzikowski, line 89: changes in package log seem redundant - there were no package.json changes. It looks like package-log was updated by using older (?) node.js version. My suggestion is to revert this file

umegbewe (approved):

Issue #647, opened 2026-01-18
Fix test `test-04-v3-snapshot-ccaas.sh`, which fails probably due to race condition


dzikowski:
An additional sleep before the relevant chaincode call might be the valid workaround

PR #648, opened 2026-01-18
test: fix flaky snapshot ccaas test with retry loop

OsamaRab3:
fix: #647 
The test passed locally but was flaky in CI due to a race condition.
Added a retry with a short delay before the first REST invoke


e2e-network/docker/test-04-v3-snapshot-ccaas.sh
```diff
@@ -122,9 +122,15 @@ echo "$user_token_response"
 user_token="$(echo "$user_token_response" | jq -r '.token')"
 
 # check if state is kept after restoration
-expectInvokeRest "$fablo_rest_org1 $user_token" "my-channel1" "chaincode1" \
-  "KVContract:get" '["name"]' \
-  '{"response":{"success":"Mr Freeze"}}'
+sleep 10
+for i in {1..10}; do
```
OsamaRab3: The test is now passing consistently, so a retry loop is no longer required

PR #649, opened 2026-01-21
Fix: Get chaincode container name by peer address instead of index

dzikowski:


e2e/__snapshots__/fabloCommands.test.ts.snap
```diff
@@ -2492,6 +2493,7 @@ exports[`extend config should extend default config 1`] = `
 "Validation errors count: 0
 Validation warnings count: 0
 ===========================================================
+Could not check for updates. Url: 'https://api.github.com/repos/hyperledger-labs/fablo/releases' not available
```
dzikowski:
```suggestion
```

PR #684, opened 2026-04-04
fix: chaincode-functions.sh always resolved to v2 template for Fabric v3 networks

### LINKED ISSUE #683: chaincode-functions.sh always gets the v2 template even on Fabric v3 networks
Was going through the source and noticed something off in src/setup-docker/index.ts, inside _copyUtilityScripts() around line 282:

- const chaincodeFunctionsTemplate = getTemplatePath(
-   this.templatesDir,
-   `fabric-docker/scripts/chaincode-functions-${capabilities.isV2 ? "v2" : "v2"}.sh`,
- );

Both sides of the ternary say "v2" — so no matter what Fabric version you're using, chaincode-functions.sh always gets generated from the v2 template.

The other scripts in the same method do it correctly:

- // these are fine
- `fabric-docker/scripts/cli/channel_fns-${capabilities.isV3 ? "v3" : "v2"}.sh`
- `fabric-docker/scripts/base-functions-${capabilities.isV3 ? "v3" : "v2"}.sh`

So the pattern was right everywhere else, just this one was missed.

Fix should be:

- `fabric-docker/scripts/chaincode-functions-${capabilities.isV3 ? "v3" : "v2"}.sh`
-

Also, there's currently no chaincode-functions-v3.sh in the templates directory — only chaincode-functions-v2.sh exists. So either a v3 variant needs to be created, or if there are no differences yet, at least the selection logic should be correct so future changes can diverge properly.

No visible error is thrown right now because the v2 file exists at both paths, but Fabric v3 users are silently getting the wrong script.

Happy to open a PR for this if it looks good.

Shivampal157:
Fixes #683

The ternary in _copyUtilityScripts() was:

    capabilities.isV2 ? "v2" : "v2"

Both branches returned "v2", so chaincode-functions.sh was always 
generated from the v2 template regardless of Fabric version. Fixed it 
to use isV3 like the other scripts in the same method do.

Also added chaincode-functions-v3.sh (mirrors v2 for now) so the 
selection actually has something to resolve to, and v3-specific changes 
can be added there later without touching v2.

### DIFF
--- a/src/setup-docker/index.ts
+++ b/src/setup-docker/index.ts
@@ -279,7 +279,7 @@ export default class SetupDocker extends Command {
     // Copy chaincode-functions script
     const chaincodeFunctionsTemplate = getTemplatePath(
       this.templatesDir,
-      `fabric-docker/scripts/chaincode-functions-${capabilities.isV2 ? "v2" : "v2"}.sh`,
+      `fabric-docker/scripts/chaincode-functions-${capabilities.isV3 ? "v3" : "v2"}.sh`,
     );
     const chaincodeFunctionsDest = getDestinationPath(this.outputDir, "fabric-docker/scripts/chaincode-functions.sh");
     await renderTemplate(chaincodeFunctionsTemplate, chaincodeFunctionsDest, {});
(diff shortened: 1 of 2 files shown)


dzikowski:
`v2` in this context does not mean the fabric version that is used, but the version of chaincode installation process (system channel vs no system channel). The process as far as know remains the same for Fabric v3, so we don't need to duplicate that code.

PR #695, opened 2026-04-10
build: bump actions and enable multi-arch support

### LINKED ISSUE #694: Enhance CI/CD pipeline for multi-platform support and build reliability
## Description:
The current GitHub Actions workflow has a few limitations that hinder development speed and cross-platform compatibility:
* **Architecture:** We currently only target `amd64`, making it difficult for users on ARM-based systems (like Apple Silicon) to use the images natively.
* **Dependency Management:** The `npm install` step is fragmented between the YAML and the build script.
* **Tooling:** Version parsing relies on `jq`, which may not be present in every environment.

## Proposed Changes:
1. Integrate `docker/setup-qemu-action` to allow cross-compilation.
2. Update all GitHub Actions to their latest stable versions (v3/v4).
3. Centralize build logic within `fablo-build.sh` using `npm ci` for consistent builds.
4. Use `node` to parse `package.json` versions to reduce external dependencies.

HarK-github:
## Overview:
This PR updates the GitHub Actions workflow to ensure more robust, cross-platform Docker builds. The primary goal was to enable `arm64` support while cleaning up how dependencies and versions are handled.

## Key Changes:
* **Multi-Platform Support:** Added **QEMU** setup to the workflow. This allows `docker buildx` to successfully build the `linux/arm64` slice of the image on GitHub's x86_64 runners.
* **Reliable Installs:** Switched to `npm ci` inside the build script. This ensures that the exact versions in `package-lock.json` are respected, preventing "dependency drift" during CI.
* **Reduced Overhead:** Removed the redundant `npm install` from the `.yml` file, as the build script now handles its own environment setup (including `nvm` support).
* **Action Upgrades:** Updated `checkout` to `v4` and `setup-buildx` to `v3` for better performance and security.
* **Version Parsing:** Switched from `cat | jq` to a native `node` call for extracting the project version, making the script more portable.

## Testing:
- Verified that `fablo-build.sh --push` triggers the multi-platform build.

- Confirmed that `shellcheck` and `yamllint` are correctly installed and executed in the runner.

**Does this require a documentation update?**
No, the public interface of the build script remains the same.

### DIFF
--- a/.github/workflows/publish-docker-on-main.yml
+++ b/.github/workflows/publish-docker-on-main.yml
@@ -11,12 +11,20 @@ jobs:
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
+
+      - name: Install lint tools
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y shellcheck yamllint
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
         with:
           driver: docker-container
+
+      - name: Set up QEMU for multi-platform builds
+        uses: docker/setup-qemu-action@v3
 
       - name: Login to GHCR
         uses: docker/login-action@v3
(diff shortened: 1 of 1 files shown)


dzikowski (commented):
This PR adds no actual value for Fablo CI processes. There are some updates that are valid, like version bump, but mostly it adds an overhead - switching from jq to node and installing redundant qemu.

PR #698, opened 2026-04-11
fix: prevent fabloRest and peerDevMode from being used simultaneously (#643)

### LINKED ISSUE #643: Using dev mode makes Fablo REST unusable
Trying to query or invoke using Fablo REST will result in a `400 Bad Request` with a messages like:
`message: 'Query failed. Errors: ["Error: Peer peer0.org1.example.com:7041 is not running chaincode myChaincode"]'` for query, or
`message: 'DiscoveryService: myChaincode error: failed constructing descriptor for chaincodes:<name:"myChaincode" > '` for invoke

I believe that this is due to dev mode peers not exposing connected chaincodes through the discovery service.

If no fix or workaround is possible, I believe a warning or error about this should be thrown when generating or running a configuration that uses both features.

zyzzmohit:
Resolves #643.

### Description
Currently, when a Fablo configuration attempts to use `global.peerDevMode: true` whilst an organization uses `tools.fabloRest: true`, the Fablo REST server will crash attempting to use discovery services, throwing a 400 Bad Request since dev mode peers don't expose connected chaincodes.

This PR introduces an explicit validation check into `src/commands/validate/index.ts` during the core CLI validation phase to gracefully reject this configuration. If `peerDevMode` is active globally, it checks if any org tries to mount `fabloRest` and properly emits a native configuration `validationErrorType.ERROR`. This ensures developers are made visually aware of the incompatibility immediately inside the CLI output during `fablo generate`, rather than finding out via REST runtime failures.

### Changes Made
*   Added `_validateFabloRestAndDevMode` module constraint in `src/commands/validate/index.ts`.
*   Integrated constraint properly into the primary `async validate()` lifecycle.
*   Formatting verified via `npm run lint` & `npm run build`.

### Testing Verification (Per Guidelines)
- [x] Executed **Unit tests** successfully (`npm run test:unit`).
- [x] Executed **End-to-End (E2E) tests** successfully (`npm run test:e2e`).
- [x] No templates/snapshots were altered during this TS logic fix, so snapshot updates (`test:e2e-update`) were safely bypassed.

Signed-off-by: zyzzmohit <mohitray949@gmail.com>


dzikowski (approved):

zyzzmohit:
@dzikowski I just clicked 'Update branch' to keep my branch synced with `main` to avoid any conflicts, but it seems that action paused the workflows again since it created a new merge commit

Could you please re-approve the workflow runs so we can get the green checks on this latest commit? Thank you :)

PR #680, opened 2026-04-02
fix(ccaas): avoid package overwrite by peer-specific archive name

Sukuna0007Abhi:
Fixes #653  by preventing CCaaS chaincode package overwrite across peers.

This PR updates CCaaS packaging/install logic to use a peer-specific archive name (${CHAINCODE_LABEL}_${peer}.tar.gz) instead of a shared ${CHAINCODE_LABEL}.tar.gz. It also adds install-time fallback to the legacy shared filename for backward compatibility.

PR #699, opened 2026-04-11
fix(ccaas): use peer-specific archive name to prevent package overwrite

### LINKED ISSUE #653: Bug: CCaaS uses only the last chaincode
In CCaaS setup eventually all chaincode requests end up calling the last CCaaS container, because the chaincode connection file (`my-channel1_chaincode1_0.0.1.tar.gz`) is overwritten. The chaincode connection file name should probably contain peer name.

More details: https://github.com/hyperledger-labs/fablo/issues/644#issuecomment-3739256651

As a workaround in the current version you can use the following `postGenerate` hook:

```
  "hooks": {
    "postGenerate": "perl -i -pe 's/\\$CHAINCODE_LABEL\\.tar\\.gz/\\${CHAINCODE_LABEL}_\\${PEER_ADDRESS%:*}.tar.gz/g' \"./fablo-target/fabric-docker/scripts/chaincode-functions.sh\""
  }
```

hellokitty09:
In CCaaS setups with multiple peers, the chaincode connection package ($CHAINCODE_LABEL.tar.gz) was being overwritten during the packaging loop because the filename didn't include the peer name. This caused all peers to install the last package, routing all chaincode requests to the last CCaaS container.

Changes:
- chaincodePackageCCaaS(): archive name now includes peer name (${CHAINCODE_LABEL}_${PEER_NAME}.tar.gz)
- chaincodeInstall(): detects peer-specific package if available, falls back to shared package for non-CCaaS chaincodes

TLS support is fully preserved.

Fixes #653

### DIFF
--- a/e2e/__snapshots__/fablo-config-hlf2-1org-1chaincode-peer-dev-mode.json.test.ts.snap
+++ b/e2e/__snapshots__/fablo-config-hlf2-1org-1chaincode-peer-dev-mode.json.test.ts.snap
@@ -2590,6 +2590,7 @@ chaincodePackageCCaaS() {
   local CHANNEL_NAME=$9
   local ORG_DOMAIN=\${10}
   local CHAINCODE_LABEL="\${CHANNEL_NAME}_\${CHAINCODE_NAME}_$CHAINCODE_VERSION"
+  local PEER_NAME="\${PEER_ADDRESS%:*}"
 
   echo "Packaging CCaaS chaincode $CHAINCODE_NAME..."
   inputLog "CHAINCODE_VERSION: $CHAINCODE_VERSION"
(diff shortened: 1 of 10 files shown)


joaormatos:
Hi,

This PR is attempting to solve the same problem as PR #680, which initially had a similar approach to this one — one package per peer — but has since been revised to address the fundamental issue with that approach that I described here: https://github.com/hyperledger-labs/fablo/issues/653#issuecomment-4192939495

Also note the ensuing discussion about the distinction between:

A) using TLS at the network level, which is an important feature to maintain, and

B) using TLS in Peer-to-CCaaS connections, which is independent of the above and has very limited usefulness in typical Fablo use cases while introducing significant complexity and overhead.

PR #680 should currently fix issue #653 while preserving A and removing B.

Having optional support for B would be nice, but it may not be worth it considering the development and maintenance cost.

joaormatos:
Hi,

This PR is attempting to solve the same problem as PR #680, which initially had a similar approach to this one — one package per peer — but has since been revised to address the fundamental issue with that approach that I described here: https://github.com/hyperledger-labs/fablo/issues/653#issuecomment-4192939495

Also note the ensuing discussion about the distinction between:

A) using TLS at the network level, which is an important feature to maintain, and

B) using TLS in Peer-to-CCaaS connections, which is independent of the above and has very limited usefulness in typical Fablo use cases while introducing significant complexity and overhead.

PR #680 should currently fix issue #653 while preserving A and removing B.

Having optional support for B would be nice, but it may not be worth it considering the development and maintenance cost.

PR #701, opened 2026-04-11
feat: Add Engine-specific Validation Rules for Kubernetes Defaults

ARYANPATEL-BIT:
## Summary

Expand `_validateEngineSpecificSettings(networkConfig: FabloConfigJson)` so Fablo validates deployment-engine-specific configuration earlier during `fablo validate`, instead of letting invalid Docker or Kubernetes setups fail later during `generate` or `up`. The current code only blocks Kubernetes when `global.fabricVersion < 2.0.0`, while the config model already distinguishes engines through `global.engine: "kubernetes" | "docker"`.

## Problem

At the moment, engine-specific validation is incomplete. The validator only checks one Kubernetes rule:

```ts
_validateEngineSpecificSettings(networkConfig: FabloConfigJson): void {
    if (networkConfig.global.engine === "kubernetes") {
      if (!version(networkConfig.global.fabricVersion).isGreaterOrEqual("2.0.0")) {
        this.emit(validationErrorType.ERROR, {
          category: validationCategories.GENERAL,
          message: `Kubernetes is not supported by Fablo for Fabric below version 2.0.0`,
        });
      }
    }
    // TODO engine-specific validation rules
  }
```

This means users can still provide engine-incompatible configuration that passes `validate` but fails later in runtime-oriented commands.

## Goal

Implement a clear rule set for both supported engines and enforce those rules inside `_validateEngineSpecificSettings`. The validator should emit:

- `ERROR` for invalid or conflicting settings that make a network definition unusable for the selected engine.
- `WARN` for settings that are technically allowed in JSON but likely ignored, misleading, or unsupported for that engine.

## Scope

The config model shows that Fablo currently supports exactly two deployment engines through `global.engine`: `docker` and `kubernetes`. Validation should therefore branch only on those two values and inspect related parts of the network definition such as global settings, org/peer/orderer exposure, tooling, and chaincode path assumptions.

### Docker-specific rules

Validate settings that are meaningful for local/container runtime execution and reject Kubernetes-only assumptions.

Suggested checks:
- Reject Kubernetes-only deployment assumptions when `global.engine === "docker"`.
- Ensure all components that must be reachable in Docker have the required local/container address or port data populated consistently.
- Warn when config contains values that are harmless in JSON but are only meaningful in Kubernetes manifests or Helm-style deployment flows.
- Validate that chaincode and tooling settings expected to work from local/container paths are present and not defined in a cluster-only way.

Suggested examples of concrete Docker rules:
- `ERROR` if a required exposed port for a peer, CA, orderer, Explorer, or REST tool is missing while Docker deployment depends on host/container port mappings.
- `WARN` if Kubernetes-oriented overrides or deployment metadata are provided but will not affect Docker execution.
- `ERROR` if a required local path such as chaincode directory resolution cannot work with Docker-based generation/execution.

### Kubernetes-specific rules

Build on the existing Fabric-version rule and validate cluster-oriented constraints more comprehensively.

Suggested checks:
- Keep the existing `Fabric >= 2.0.0` requirement for Kubernetes.
- Reject Docker-only assumptions that rely on host-bound port exposure where Kubernetes networking should use service-based access.
- Warn or error on settings that imply local-container behavior rather than Kubernetes service discovery.
- Validate that any required Kubernetes-compatible addresses, domains, or exposure assumptions are consistent across orgs, peers, CAs, orderers, and tools.

Suggested examples of concrete Kubernetes rules:
- `ERROR` if Kubernetes is selected with Fabric lower than `2.0.0`.
- `WARN` if explicit host-exposed ports are configured but are not meaningful or are ignored under Kubernetes deployment mode.
- `ERROR` if component addressing depends on Docker-style direct host/container networking rather than Kubernetes-compatible service naming assumptions.
- `WARN` if optional tools are enabled with settings that Docker may tolerate but Kubernetes deployment cannot meaningfully honor.

### Target method

Implement the new checks in:

`src/commands/validate/index.ts`

Main target:

```ts
_validateEngineSpecificSettings(networkConfig: FabloConfigJson): void
```

Recommended structure:
1. Read `const engine = networkConfig.global.engine`.
2. Dispatch to two private helpers for readability, for example:
   - `_validateDockerSpecificSettings(networkConfig)`
   - `_validateKubernetesSpecificSettings(networkConfig)`
3. Reuse existing `this.emit(validationErrorType.ERROR | WARN, ...)` patterns and existing validation categories where possible.
4. Keep messages explicit about:
   - which engine was selected,
   - which field or component is invalid,
   - what the user should change.

### Validation style

Prefer deterministic field-level checks over broad heuristic warnings. A good rule should identify:
- the engine,
- the offending setting,
- why it conflicts,
- whether it blocks execution (`ERROR`) or is advisory (`WARN`).

Example message style:

```ts
this.emit(validationErrorType.ERROR, {
  category: validationCategories.GENERAL,
  message: `Setting "<field>" is not valid for engine "${networkConfig.global.engine}" because ...`,
});
```

## Tests

Add unit tests in:

`src/commands/validate/index.test.ts`

Tests should cover both valid and invalid configurations so the validate phase catches engine mismatches before `generate` or `up`.

#### Docker
- Passes for a valid Docker network configuration.
- Emits `ERROR` for each required Docker-only field that is missing or inconsistent.
- Emits `WARN` for Kubernetes-only or ignored settings present in Docker mode.

#### Kubernetes
- Emits `ERROR` when Fabric version is below `2.0.0`.
- Passes for a valid Kubernetes configuration on supported Fabric versions.
- Emits `ERROR` for Docker-only assumptions that are incompatible with Kubernetes mode.
- Emits `WARN` for settings that are accepted in config but ignored or misleading in Kubernetes mode.

### Test design guidance
- Start from existing valid fixtures and mutate one field at a time.
- Assert both severity and message text.
- Keep one rule per test where possible, so failures identify exactly which validation broke.
- Add at least one “happy path” test per engine to avoid over-restrictive validation.

## Acceptance criteria
- Identify required and conflicting configuration parameters specific to `docker`.
- Identify required and conflicting configuration parameters specific to `kubernetes`.
- Implement these checks inside `_validateEngineSpecificSettings`.
- Emit appropriate `ERROR` and `WARN` results through the existing validation mechanism.
- Add unit tests covering valid and invalid configurations for both engines.

## Suggested PR title

`feat(validate): add engine-specific validation rules for docker and kubernetes`

## Suggested commit message

```bash
feat(validate): enforce engine-specific config rules
```


ARYANPATEL-BIT:
@dzikowski , @OsamaRab3, @umegbewe

please review this pr

PR #712, opened 2026-04-13
 Validate CCaaS chaincode port limit

ARYANPATEL-BIT:
## Description
This PR addresses an edge case in CCaaS (Chaincode-as-a-Service) networking where the calculated chaincode container port could exceed the maximum valid port limit (65535).

Previously, the port logic `10000 * (index + 1) + peer.port` would silently allocate invalid ports (e.g., `77021` for the 7th chaincode index), causing downstream cluster deployment failures.

This fix introduces a proper boundary check during configuration generation that will aggressively validate the port limitation and throw a helpful, descriptive error to the user during the initial `$ fablo generate` or `$ fablo up` phase, saving time and preventing confusing orchestrator crashes.

**Resolves Issue:** Fixes #711 

## Changes Made
- Added a validation guard in `src/extend-config/extendChaincodesConfig.ts`
- The logic now verifies that the calculated CCaaS port does not exceed `65535`
- Throws a descriptive `Error` explicitly guiding the user on how to avoid the limitation

## Testing Performed
- Verified code mathematically protects against ports > 65535
- Verified existing snapshot and unit tests continue to pass (`npm run test:unit`)

### DIFF
--- a/src/extend-config/extendChaincodesConfig.ts
+++ b/src/extend-config/extendChaincodesConfig.ts
@@ -86,10 +86,17 @@ const extendChaincodesConfig = (
             const containerName = `ccaas_${peer.address.replace(/[^a-zA-Z0-9_.-]/g, "_")}_${channel.name}_${
               chaincode.name
             }_${versionSuffix}`.toLowerCase();
+            const port = 10000 * (index + 1) + peer.port;
+            if (port > 65535) {
+              throw new Error(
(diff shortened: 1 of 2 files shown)


ARYANPATEL-BIT:
Hi @umegbewe , @dzikowski

Could you please review this PR when you get a chance? I’d really appreciate your feedback.

Also, let me know if you’d like me to make any additional changes or improvements—I’m happy to iterate on it.

Thanks in advance!

PR #731, opened 2026-04-22
fix: support peerDevMode with TLS enabled

### LINKED ISSUE #573: Support dev mode with TLS
To be checked if it is sufficient to provide TLS envs for the chaincode, like here:

[CORE_CHAINCODE_TLS_CERT_FILE](https://hyperledger-fabric.readthedocs.io/en/release-2.2/cc_service.html)

> If TLS is enabled, the following additional options are required:
>
> - CORE_CHAINCODE_TLS_CERT_FILE (–chaincode-tls-cert-file): path to a certificate
> - CORE_CHAINCODE_TLS_KEY_FILE (–chaincode-tls-key-file): path to a private key

Abhishek9639:
Fixes #573

Previously, using `peerDevMode: true` and `tls: true` together in the config would throw a validation error and block network generation. But this combination is actually valid Hyperledger Fabric supports it as long as the peer has the right chaincode TLS environment variables set.

**What changed:**
- Removed the validation error that was blocking `peerDevMode + tls` combination
- Added `CORE_CHAINCODE_TLS_CERT_FILE` and `CORE_CHAINCODE_TLS_KEY_FILE` env vars to the peer container in the docker-compose template (only injected when both `peerDevMode` and `tls` are enabled)
- Added a sample config demonstrating the working combination
- Added an E2E test for the new sample

**Testing:**
- All existing unit tests pass (34/34)
- All E2E snapshot/schema tests pass including the new one (50/50)
- New sample config passes schema validation

### DIFF
--- a/e2e/__snapshots__/fablo-config-hlf2-1org-1chaincode-peer-dev-mode-tls.json.test.ts.snap
+++ b/e2e/__snapshots__/fablo-config-hlf2-1org-1chaincode-peer-dev-mode-tls.json.test.ts.snap
@@ -1,0 +1,9 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`samples/fablo-config-hlf2-1org-1chaincode-peer-dev-mode-tls.json should create proper e2e/__tmp__/samples/fablo-config-hlf2-1org-1chaincode-peer-dev-mode-tls.json.tmpdir/fablo-target/fablo-config.json from samples/fablo-config-hlf2-1org-1chaincode-peer-dev-mode-tls.json 1`] = `""`;
+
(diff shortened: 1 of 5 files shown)


Abhishek9639:
Hi @dzikowski,
could you please take a look at this when you get a chance? 

This fixes #573 the validation was blocking peerDevMode and TLS from being used together, even though Fabric actually supports it. The fix is pretty minimal just removed the validation error and added the two required chaincode TLS env vars to the peer container in the docker-compose template.

All existing tests still pass. Let me know if you'd like any changes.
Thanks

dzikowski:
Please run locally a test with peer dev mode enabled, and TLS enabled. I believe it wont work, because the fundamental issue seems not resolved. See #573

dzikowski:
it seems tests are failing, because snapshots are not updated. Take a look at: https://github.com/hyperledger-labs/fablo/blob/main/CONTRIBUTING.md

Abhishek9639:
Ok, I’ll check it.

PR #733, opened 2026-04-23
docs: add WSL2 troubleshooting guide to README

shahedalhunity754:
Added specific steps for socket permissions and docker compose syntax for WSL2 users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,14 @@ Note: If you install Fablo as a local script, you call it as `./fablo <command>`
 If you install it globally, you call `fablo <command>`.
 For the simplicity we will refer to it as `fablo`.
 
+###Troubleshooting for WSL2 & Ubuntu Users 
+If you are running Fablo on **Windows Subsystem for Linux (WSL2)**, you may encounter environment-specific issues. Here are the recommended fixes:
+
+* **Docker V2 Syntax:** Ensure you use `docker compose` (with a space) instead of `docker-compose`.
+* **Socket Permissions: **If you get a "permission denied" error, run:
+   `sudo chmod 666 /var/run/docker.sock`
+* **command Stability:** In WSL2, use the full flag `--detach` instead of `-d` when intializing the network to ensure the containers start correctly.
+* **WSL2 Integration:** Confirm that "WSL2 Integration" is enabled in Docker Desktop settings for your specific Ubuntu distribution. 
 
 ## Basic usage
 
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ If you are running Fablo on **Windows Subsystem for Linux (WSL2)**, you may enco
 * **Docker V2 Syntax:** Ensure you use `docker compose` (with a space) instead of `docker-compose`.
 * **Socket Permissions: **If you get a "permission denied" error, run:
    `sudo chmod 666 /var/run/docker.sock`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -65,6 +65,14 @@ Note: If you install Fablo as a local script, you call it as `./fablo <command>`
 If you install it globally, you call `fablo <command>`.
 For the simplicity we will refer to it as `fablo`.
 
+###Troubleshooting for WSL2 & Ubuntu Users 
+If you are running Fablo on **Windows Subsystem for Linux (WSL2)**, you may encounter environment-specific issues. Here are the recommended fixes:
+
+* **Docker V2 Syntax:** Ensure you use `docker compose` (with a space) instead of `docker-compose`.
+* **Socket Permissions: **If you get a "permission denied" error, run:
+   `sudo chmod 666 /var/run/docker.sock`
+* **command Stability:** In WSL2, use the full flag `--detach` instead of `-d` when intializing the network to ensure the containers start correctly.
```
umegbewe: The --detach vs -d point shouldn’t be necessary here. Both flags are equivalent, and Fablo abstracts container startup anyway, so users shouldn’t need to handle this manually.

README.md
```diff
@@ -65,6 +65,14 @@ Note: If you install Fablo as a local script, you call it as `./fablo <command>`
 If you install it globally, you call `fablo <command>`.
 For the simplicity we will refer to it as `fablo`.
 
+###Troubleshooting for WSL2 & Ubuntu Users 
+If you are running Fablo on **Windows Subsystem for Linux (WSL2)**, you may encounter environment-specific issues. Here are the recommended fixes:
+
+* **Docker V2 Syntax:** Ensure you use `docker compose` (with a space) instead of `docker-compose`.
+* **Socket Permissions: **If you get a "permission denied" error, run:
```
umegbewe, line 72: It would be good to confirm the exact scenario where this was required, did you experience this error?

README.md
```diff
@@ -65,6 +65,14 @@ Note: If you install Fablo as a local script, you call it as `./fablo <command>`
 If you install it globally, you call `fablo <command>`.
 For the simplicity we will refer to it as `fablo`.
 
+###Troubleshooting for WSL2 & Ubuntu Users 
+If you are running Fablo on **Windows Subsystem for Linux (WSL2)**, you may encounter environment-specific issues. Here are the recommended fixes:
+
+* **Docker V2 Syntax:** Ensure you use `docker compose` (with a space) instead of `docker-compose`.
```
umegbewe, line 71: The docker compose vs docker-compose distinction is not specific to WSL2. That’s simply the difference between Compose v2 (Docker CLI plugin) and the older standalone v1, and applies across all environments.

shahedalhunity754: Hi @umegbewe , thank you for your insightful feedback! 
Regarding your points:
1. I'll remove the --detach note to keep the guide focused on essentials fixes.
2. For Socket Permissions, I encountered the 'permission denied' error on a fresh WSL2 setup, and that command was the only way to get the Fablo running.
3. You're right, the Docker V2 syntax is a general point, but I added it because many new WSL2 users often rely on outdated tutorials.
I will update the PR and push the changes shortly. Is there anything else you think I should add or modify to make this guide more helpful?
Thanks again!

PR #736, opened 2026-04-28
fix: add retry with || true for post-snapshot invoke in test-04

### LINKED ISSUE #734: test-04-v3-snapshot-ccaas still flaky - sleep insufficient, needs retry with || true pattern
## Description

`test-04-v3-snapshot-ccaas.sh` fails intermittently in CI. The first `expectInvokeRest` call after snapshot restore (line 132) hits `DEADLINE_EXCEEDED` due to a race condition between the CCaaS container being ready and the peer re-establishing its gRPC connection.

This was previously addressed in #647 → #648, but the fix (`sleep 2`, later `sleep 3`) is insufficient under variable CI load. 

---

## Root Cause

[`waitForContainer`](https://github.com/hyperledger-labs/fablo/blob/main/e2e-network/docker/test-04-v3-snapshot-ccaas.sh#L120-L121) confirms the CCaaS container logged `"Bootstrap process completed"` – meaning the gRPC **server** is listening. However, the Fabric **peer** re-establishing its gRPC **client** connection happens asynchronously after that log message. A fixed sleep cannot reliably account for this under variable CI load.

---
## Why `sleep` Alone Isn't Sufficient (and why the original retry was tricky)

PR #648 added a `sleep` before the invoke, which helps in most cases.The PR also originally explored a retry loop, but there's a subtle interaction with the test environment that made it ineffective:

1. [`test-04` uses `set -e`](https://github.com/hyperledger-labs/fablo/blob/main/e2e-network/docker/test-04-v3-snapshot-ccaas.sh#L3)  – aborts on any non-zero exit
2. [`expectInvokeRest`](https://github.com/hyperledger-labs/fablo/blob/main/e2e-network/docker/test-04-v3-snapshot-ccaas.sh#L38-L40) calls `expect-invoke-rest.sh` via `sh`
3. [`expect-invoke-rest.sh` calls `exit 1` on failure](https://github.com/hyperledger-labs/fablo/blob/main/e2e-network/docker/expect-invoke-rest.sh#L54)  – hard exit
4. Under `set -e`, the non-zero exit code aborts the entire script before the loop can retry

The `|| true` pattern from `test-01` addresses this by absorbing the exit code, allowing the loop to continue to the next attempt.

---

## Proven Solution Already in the Codebase

[`test-01-v3-simple.sh` lines 48–61](https://github.com/hyperledger-labs/fablo/blob/main/e2e-network/docker/test-01-v3-simple.sh#L48-L61) solve this exact class of problem with `expectQueryWithRetry`:

```bash
output="$(expectQuery "$1" "$2" "$3" "$4" "$5" 2>&1 || true)"
```

- `2>&1` captures both stdout and stderr into the variable
- `|| true` absorbs the non-zero exit code so `set -e` doesn't abort
- `grep` checks the captured output for the failure marker

---

## Proposed Fix

Add `expectInvokeRestWithRetry` using the same pattern:

```bash
expectInvokeRestWithRetry() {
  local output=""
  for i in $(seq 1 10); do
    output="$(expectInvokeRest "$@" 2>&1 || true)"
    if echo "$output" | grep -qF "ok (rest)"; then
      echo "$output"
      return 0
    fi
    echo "Invoke failed after restore, retrying... ($i/10)"
    sleep 3
  done
  echo "$output"
  echo "❌ All retries exhausted"
  exit 1
}
```

We'll apply only to the **first** `expectInvokeRest` after snapshot restore (line 132). The second invoke runs normally since the connection is already established by then.

---

Happy to open a PR for this fix if it sounds good.

## Related

- #647  : original issue (closed)
- #648  : previous fix with `sleep` (merged, insufficient)
- #728  :  re-confirmed as known issue by maintainer

Slambot01:
Fixes #734 

The first `expectInvokeRest` after snapshot restore intermittently fails with `DEADLINE_EXCEEDED`. `waitForContainer` confirms the CCaaS gRPC server is listening, but the peer's gRPC client reconnection happens asynchronously - `sleep 2` (from #648) is insufficient under variable CI load.

The previous retry attempt in #648 used `&& break` without `|| true`, which couldn't work because `expect-invoke-rest.sh` calls `exit 1` on failure and the test runs under `set -e`, aborting before the loop can retry.

### Fix

Added `expectInvokeRestWithRetry` using the same `|| true` pattern proven in `test-01-v3-simple.sh` (`expectQueryWithRetry`, lines 48–61). Applied only to the first invoke after snapshot restore .the second invoke runs normally since the connection is established by then.


dzikowski (dismissed):

dzikowski:
I initially approved, but then I realized it does not solved the underlying issue. Test still fail, probably because the error is in a different place. Not in the test, and waiting till the container is ready - but somewhere in the network boot/restore process. There is probably race condition when the chaincode container starts. Maybe as a workaround we should try to restart it in the test to verify if that's the issue? And then fix in a proper way, eliminating the root cause.

Slambot01:
> I initially approved, but then I realized it does not solved the underlying issue. Test still fail, probably because the error is in a different place. Not in the test, and waiting till the container is ready - but somewhere in the network boot/restore process. There is probably race condition when the chaincode container starts. Maybe as a workaround we should try to restart it in the test to verify if that's the issue? And then fix in a proper way, eliminating the root cause.

Makes sense. I’ll look into the chaincode container restart angle. I’ll check if restarting the CCaaS container after snapshot restore fixes the failure, and if it does, I’ll trace it back to the actual issue in the restore flow.

Slambot01:
> I initially approved, but then I realized it does not solved the underlying issue. Test still fail, probably because the error is in a different place. Not in the test, and waiting till the container is ready - but somewhere in the network boot/restore process. There is probably race condition when the chaincode container starts. Maybe as a workaround we should try to restart it in the test to verify if that's the issue? And then fix in a proper way, eliminating the root cause.

The CCaaS restart approach worked locally, but CI is still failing. All 10 retries end up hitting `DEADLINE_EXCEEDED`.
The CCaaS containers are bootstrapping properly after restart, and the peer gRPC state is showing `READY`, so it doesn’t look like the issue is with the CCaaS ↔ peer connection itself.At this point the problem seems to be somewhere deeper in the restore flow. I need to dig into the state the peer or fablo-rest ends up in after restore, because something there is preventing the endorsement flow from completing properly. I guess ,will need to see further for this.

PR #756, opened 2026-05-13
fix: use headPeer.address in anchor peer update instead of hardcoded peer0

### LINKED ISSUE #751: Anchor peer update uses hardcoded peer0 prefix, ignores custom peer prefix config
## Summary

`notifyOrgAboutNewChannel` and `notifyOrgAboutNewChannelTls` in `commands-generated.sh` are hardcoding `peer0.<org.domain>` as the peer address instead of using the `peer.prefix` from the config. Everything else in the same file handles the prefix correctly, so this one just got missed.

You can see it clearly if you run:
1. Set up a config with `"peer": { "instances": 2, "prefix": "node" }`
2. Run `fablo generate fablo-config.json`
3. Check: `grep "notifyOrgAboutNewChannel" fablo-target/fabric-docker/commands-generated.sh`

All other operations will show `node0.org1.example.com` but `notifyOrgAboutNewChannel` is still pointing at `peer0.org1.example.com`.

## Root Cause

Lines 219 and 226 in `src/setup-docker/templates/fabric-docker/commands-generated.sh` have the peer address hardcoded:

```
"peer0.<%= org.domain %>"
```

It should be `org.headPeer.address` which is already prefix-aware. `org.headPeer` is computed from the peer list in `extendOrgsConfig.ts:258` (`headPeer: peers[0]`) and the exact same pattern is already being used correctly in `chaincode-install-v2.sh` (lines 33, 66, 89, 90) and `chaincode-dev-v2.sh` (lines 11, 26, 34).

## Impact

Any network with a custom peer prefix will fail at anchor peer update. The CLI looks for `peer0.org1.example.com` which doesn't exist as a container, Docker DNS resolution fails, and the error gives no hint that the prefix config is being ignored. Took a while to track down.

## Suggested Fix

Two lines need updating in `src/setup-docker/templates/fabric-docker/commands-generated.sh`:

```diff
-         "peer0.<%= org.domain %>" <% -%>
+         "<%= org.headPeer.address %>" <% -%>
```

Line 219 is the non-TLS case, line 226 is the TLS variant.

Opened PR for this fix.

Slambot01:
## Summary
`notifyOrgAboutNewChannel` and `notifyOrgAboutNewChannelTls` in `commands-generated.sh` were hardcoding `peer0.<org.domain>` as the peer address instead of respecting the `peer.prefix` config field. If you set a custom prefix like `"prefix": "node"`, everything else in the generated file picks it up and uses `node0.org1.example.com`, but the anchor peer update was still pointing at `peer0.org1.example.com` which doesn't exist as a container. Docker DNS silently fails with no obvious indication that the prefix config is being ignored.

## Fix
Switched to `org.headPeer.address` which already handles the prefix correctly. `headPeer` is just `peers[0]` from the prefix-aware peer list built in `extendOrgsConfig.ts:258`, and the same pattern is already used in `chaincode-install-v2.sh` and `chaincode-dev-v2.sh`. All 17 unit tests pass.

Fixes #751

### DIFF
--- a/e2e/__snapshots__/fablo-config-hlf2-2orgs-2chaincodes-raft.yaml.test.ts.snap
+++ b/e2e/__snapshots__/fablo-config-hlf2-2orgs-2chaincodes-raft.yaml.test.ts.snap
@@ -4314,8 +4314,8 @@ notifyOrgsAboutChannels() {
   printHeadline "Notyfing orgs about channels" "U1F4E2"
   notifyOrgAboutNewChannelTls "my-channel1" "Org1MSP" "cli.org1.example.com" "peer0.org1.example.com" "orderer0.group1.orderer1.com:7030" "crypto-orderer/tlsca.orderer1.com-cert.pem"
   notifyOrgAboutNewChannelTls "my-channel1" "Org2MSP" "cli.org2.example.com" "peer0.org2.example.com" "orderer0.group1.orderer1.com:7030" "crypto-orderer/tlsca.orderer1.com-cert.pem"
-  notifyOrgAboutNewChannelTls "my-channel2" "Org1MSP" "cli.org1.example.com" "peer0.org1.example.com" "orderer0.group1.orderer1.com:7030" "crypto-orderer/tlsca.orderer1.com-cert.pem"
-  notifyOrgAboutNewChannelTls "my-channel2" "Org2MSP" "cli.org2.example.com" "peer0.org2.example.com" "orderer0.group1.orderer1.com:7030" "crypto-orderer/tlsca.orderer1.com-cert.pem"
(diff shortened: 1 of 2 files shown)


Slambot01:
test-02-raft failure is unrelated to this PR and was already present before these changes. The CI logs show the failure happens during Java chaincode (chaincode2) installation, where build.sh crashes with "cd: null directory" even though the Gradle build completes successfully. This points to a fragile issue in the sample chaincode build script itself.The anchor peer fix in this PR only changes updateAnchorPeers inside commands-generated.sh. The channel creation logs also confirm the fix is working correctly, as my-channel2 is successfully created and joined by Org1/peer1

PR #770, opened 2026-06-23
fix: guard Consortium field for Fabric V3 in configtx.yaml

### LINKED ISSUE #757: V3 configtx.yaml renders Consortium field which was removed in Fabric V3
## Summary
`src/setup-docker/templates/fabric-config/configtx.yaml` renders `Consortium: SampleConsortium` for all Fabric versions including V3. Consortiums were removed in Fabric V3 and this field should not appear in V3 configtx output.

## Root Cause
Commit `adc995e` correctly guarded the `Consortiums` block with `!isV3`, but the scalar `Consortium: SampleConsortium` field immediately above it was left unguarded and renders for all Fabric versions including V3:

```ejs
<%_ if (!global.capabilities.isV3) { _%>
    Capabilities:
        <<: *ApplicationCapabilities
<%_ } _%>
Consortium: SampleConsortium
<%_ if (!global.capabilities.isV3) { _%>
Consortiums:
    SampleConsortium:
        Organizations:
          ...
<%_ } _%>
```

Both the `Capabilities` block above and the `Consortiums` block below are correctly gated with `!isV3`. The scalar `Consortium` line was missed.

## Impact
Every Fabric V3 network gets `Consortium: SampleConsortium` injected into the generated `configtx.yaml`. While configtxgen may currently ignore this field silently, it is incorrect output and a regression from the original intent of `adc995e`.

## Fix

```diff
-        Consortium: SampleConsortium
+        <%_ if (!global.capabilities.isV3) { _%>
+        Consortium: SampleConsortium
+        <%_ } _%>
```

Happy to open a PR for this if it looks good.

MaryamArif7:
Commit adc995e correctly guarded the Consortiums block with !isV3, but the scalar Consortium: SampleConsortium field immediately above it was left unguarded and renders for all Fabric versions including V3:

Fixes #757

### DIFF
--- a/e2e/__snapshots__/fablo-config-hlf2-1org-2chaincode-raft-ccaas.json.test.ts.snap
+++ b/e2e/__snapshots__/fablo-config-hlf2-1org-2chaincode-raft-ccaas.json.test.ts.snap
@@ -257,7 +257,6 @@ Profiles:
             <<: *Group1Defaults
             Organizations:
                 - *Orderer
-        Consortium: SampleConsortium
         Application:
             <<: *ApplicationDefaults
             Organizations:
@@ -269,7 +268,6 @@ Profiles:
             <<: *Group1Defaults
             Organizations:
                 - *Orderer
-        Consortium: SampleConsortium
         Application:
             <<: *ApplicationDefaults
             Organizations:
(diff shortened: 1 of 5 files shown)


MaryamArif7:
Applied the solution from the issue thread. Let me know changes if required

umegbewe (approved):

PR #780, opened 2026-06-30
Fix prevent ccaas package peer overwrite

### LINKED ISSUE #623: When network is recovered from snapshots, CCaaS chaincode should start automatically
Currently as our README says:

```
6. When running external chaincodes(CCAAS), Execute `./fablo chaincodes install` to start the CCAAS container
```

Probably it can be resolved by starting CCaaS containers in network start phase instead of chaincode installation, and possibly adding some retry policy on failures

### LINKED ISSUE #653: Bug: CCaaS uses only the last chaincode

dpereowei:
This update moves CCaaS workloads into the generated `docker-compose.yaml` while preserving the lifecycle ordering constraint: peers still start first, then Fablo resolves the installed package ID and starts the matching CCaaS service with `CORE_CHAINCODE_ID_NAME` / `CHAINCODE_ID`. Closes #653 

The CI failure with npm help output came from how the existing command override was rendered in Compose.
Closes #623


samples/fablo-config-hlf3-2orgs-1chaincode-raft-ccaas.json
```diff
@@ -76,10 +76,11 @@
       "version": "0.0.1",
       "lang": "ccaas",
       "channel": "my-channel1",
-      "image": "ghcr.io/fablo-io/fablo-sample-kv-node-chaincode:2.2.0"
+      "image": "ghcr.io/fablo-io/fablo-sample-kv-node-chaincode:2.2.0",
+      "chaincodeStartCommand": "npm run start:ccaas"
```
dzikowski, line 80: should we provide this also for the fablo config hardcoded for `init` command in https://github.com/hyperledger-labs/fablo/blob/main/src/commands/init/index.ts ?

src/setup-docker/templates/fabric-docker/commands-generated.sh
```diff
@@ -29,6 +29,9 @@ startNetwork() {
   printHeadline "Starting network" "U1F680"
   (cd "$FABLO_NETWORK_ROOT"/fabric-docker && docker compose up -d)
   sleep 4
+  if [ -n "$(find "$FABLO_NETWORK_ROOT/fabric-config/chaincode-packages" -type f -name '*.tar.gz' -print -quit 2>/dev/null)" ]; then
+    startCCaaSContainers false
+  fi
```
dzikowski: if I'm reading it correctly, it starts chaincodes from the existing chaincode-packages files. I'm not sure if `find` is the right check here - I'm not sure if any check is required since we get chaincode list inside `startCCaaSContainers` and do nothing if there are no chaincodes. That part of the code probably can be reduced to:


```suggestion
  startCCaaSContainers
```

src/setup-docker/templates/fabric-docker/scripts/chaincode-functions-v2.sh
```diff
@@ -194,22 +216,20 @@ chaincodeInstall() {
...
-  local CONTAINER_NAME="$8"
+  local PACKAGE_LABEL="$3"
+  local CLI_NAME="$4"
+  local CA_CERT="$5"
+  local CONTAINER_NAME="$6"
   CONTAINER_NAME=$(echo "$CONTAINER_NAME" | tr '[:upper:]' '[:lower:]')
-  local CHAINCODE_MOUNT_PATH="$9"
-  local CHAINCODE_START_COMMAND="${10}"
+  local TLS_ENABLED="$7"
+  local SKIP_IF_MISSING="${8:-false}"
```
dzikowski, line 235: under what circumstances may it happen? I'm trying to understand if it's ok to skip, or we should terminate with error in that case

samples/fablo-config-hlf3-2orgs-1chaincode-raft-ccaas.json
dpereowei: Good catch. I added chaincodeStartCommand: "npm run start:ccaas" to the hardcoded fablo init ccaas config as well.

src/setup-docker/templates/fabric-docker/scripts/chaincode-functions-v2.sh
dpereowei: The skip is intended only for network start/snapshot restore. In that path, the network may be starting before any chaincode has been installed, so there may be no package IDs to resolve yet, and skipping is safe. I kept the lower-level startCCaaSContainer fail-by-default, so explicit install/upgrade paths still fail when a package ID cannot be found.

src/setup-docker/templates/fabric-docker/commands-generated.sh
dpereowei: Agreed, simplified. startNetwork now calls startCCaaSContainers directly.

dzikowski (approved):

umegbewe (approved):

Issue #787, opened 2026-07-23
feat(fabric-x): Add experimental provider to fablo config schema and types


umegbewe:
Add a provider field without changing the behavior of existing configs.

Scope:

- Add `global.provider` to `docs/schema.json`
- Add `provider?: "fabric" | "fabric-x"` to `GlobalJson`
- Add `provider: "fabric" | "fabric-x"` to the extended `Global` type
- Default provider to `fabric` in `extendGlobal`
- Mark the schema title/description as experimental for `fabric-x`

PR #796, opened 2026-07-28
Feat788 :Add provider-aware validation gates

### LINKED ISSUE #788: feat(fabric-x): Add provider-aware validation gates
Make validation choose the right rule set before classic Fabric assumptions reject Fabric-X configs

Scope:

- In `src/commands/validate/index.ts`, derive provider early from `networkConfig.global.provider ?? "fabric"`
- Preserve the existing validation path for provider `fabric`
- Add a Fabric-X validation path for provider `fabric-x`
- Make unsupported Fabric-X settings fail fast with clear messages

Validation:
- `provider: fabric` and missing provider execute the current validation checks unchanged
- `provider: fabric-x` fails if `global.engine` is `kubernetes`
- `provider: fabric-x` fails if `global.tls` is `false`
- `provider: fabric-x` fails if `global.peerDevMode` is `true`
- `provider: fabric-x` fails if Explorer or Fablo REST is enabled
- `provider: fabric-x` fails if `chaincodes` is non-empty in the first PoC
- `provider: fabric-x` fails if there is not exactly one channel
- `provider: fabric-x` allows the PoC topology even if some classic orderer/peer rules would not apply

MaryamArif7:
fixes #788
this PR includes commits from #787 since #788 depends on it ,will rebase once that merges.

Changes

Adds provider-aware validation to `fablo validate`. When `global.provider` is 
`"fabric-x"`, a new `_validateFabricXSettings()` path runs instead of the 
classic Fabric checks, enforcing: Docker engine only, TLS required, dev mode 
off, no Explorer/Fablo REST, no chaincodes, exactly one channel.

### DIFF
--- a/src/commands/validate/index.ts
+++ b/src/commands/validate/index.ts
@@ -116,6 +116,15 @@ export default class Validate extends Command {
 
     const configContent = fs.readFileSync(this.fabloConfigPath, "utf-8");
     const networkConfig = parseFabloConfig(configContent);
+    const provider = networkConfig.global.provider ?? "fabric";
+
+    if (provider === "fabric-x") {
+      this._validateJsonSchema(networkConfig);
+      this._validateSupportedFabloVersion(networkConfig.$schema);
+      this._validateFabricXSettings(networkConfig);
+      return;
+    }
+
     this._validateFabricVersion(networkConfig.global);
     networkConfig.chaincodes.forEach((chaincode) => this._validateCcaaTLS(networkConfig.global, chaincode));
     this._validateJsonSchema(networkConfig);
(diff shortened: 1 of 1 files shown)


src/commands/validate/index.ts
```diff
@@ -116,6 +116,15 @@ export default class Validate extends Command {
...
     const networkConfig = parseFabloConfig(configContent);
+    const provider = networkConfig.global.provider ?? "fabric";
+
+    if (provider === "fabric-x") {
+      this._validateJsonSchema(networkConfig);
+      this._validateSupportedFabloVersion(networkConfig.$schema);
+      this._validateFabricXSettings(networkConfig);
+      return;
+    }
+
```
umegbewe: This is in the right direction, but is placed too early. Fabric-X still uses the shared fablo schema and still has `global.fabricVersion` so common validation should run before branching.

I'd do this in this order: parse config, validate Fabric version, validate shared schema, validate supported Fablo schema URL, then branch into Fabric-X vs classic Fabric validators

src/commands/validate/index.ts
```diff
@@ -116,6 +116,15 @@ export default class Validate extends Command {
...
     this._validateFabricVersion(networkConfig.global);
     networkConfig.chaincodes.forEach((chaincode) => this._validateCcaaTLS(networkConfig.global, chaincode));
```
umegbewe, line 129: this should stay ing the classic fabric path, we should reject chaincodes entirely in this proof of concept so avoid running ccass validations before the provider branch

umegbewe (changes requested):

src/commands/validate/index.ts
```diff
@@ -116,10 +116,17 @@ export default class Validate extends Command {
 
     const configContent = fs.readFileSync(this.fabloConfigPath, "utf-8");
     const networkConfig = parseFabloConfig(configContent);
+    const provider = networkConfig.global.provider ?? "fabric";
```
umegbewe, line 119: This is fine logically, but I’d move the provider branch after the shared validation calls. Fabric-X still uses the shared Fablo schema and fabricVersion, so should be usual validation first, Fabric-X-specific validation second, classic Fabric validation last

umegbewe: Stale comment, please ignore

umegbewe (approved):

Issue #790, opened 2026-07-23
feat(fabric-x): Route Docker setup by provider without disturbing classic Fabric


umegbewe:
Teach Docker setup to call a Fabric-X writer when `global.provider` is `fabric-x`.

Scope:

- In `src/setup-docker/index.ts`, branch after parsing/extending config
- Keep the existing writer path for provider `fabric`
- Add a provider-specific writer for Fabric-X, either as a new class or clearly isolated private methods
- Do not run classic connection profile, CA server, peer/orderer compose, chaincode, Explorer, or private data generation for this proof of concept
- Still generate hooks and the network topology only if they make sense otherwise explicitly skip or generate a Fabric-X-specific minimal topology

umegbewe:
Closed as done

PR #799, opened 2026-07-31
 Add Fabric-X lifecycle commands

MaryamArif7:
fixes #791 

What this does:

Wires up Fabric-X lifecycle commands (`up`, `start`, `stop`, `down`, `reset`, plus `namespace list`/`namespace init`) behind the existing `fablo` CLI, so a Fabric-X network can be controlled the same way classic Fabric ones already are.

- `fabric-x-docker.sh` — new generated script (template lives under `src/setup-docker/templates/`) that drives `docker compose` against the generated network, plus crypto/genesis generation and namespace setup via `fxconfig`.
- A small addition to `fablo.sh`'s `executeFabloCommand()` and `networkPrune()` so it actually dispatches to the new script when present, same pattern used for `fabric-docker.sh`/`fabric-k8s.sh`.

Command logic is based on the [fabric-x-samples devnet Makefile](https://github.com/hyperledger/fabric-x-samples/blob/main/devnet/Makefile), per the issue description.

PR #800, opened 2026-07-31
feat: route fabric-x provider through isolated FabricXDockerWriter

HarK-github:
## Description
This PR integrates routing logic to support the new `fabric-x` provider during Docker setup, while  preserving the integrity of the classic Fabric orchestration logic. 

Solves issue:  #790


src/setup-docker/fabricXDockerWriter.ts
```diff
@@ -0,0 +1,38 @@
+import { FabloConfigExtended } from "../types/FabloConfigExtended";
+import { renderTemplate, getTemplatePath, getDestinationPath } from "../utils/templateUtils";
+
+export class FabricXDockerWriter {
+  constructor(private templatesDir: string, private outputDir: string, private log: (msg: string) => void) {}
+
+  public async write(configExtended: FabloConfigExtended): Promise<void> {
+    this.log("Generating Fabric-X network files...");
+
+    const staticFiles = ["docker-compose.yaml", "crypto-config.yaml", "configtx.yaml", "shared_config.yaml"];
```
umegbewe:
```suggestion
    const staticFiles = ["docker-compose.yaml", "crypto-config.yaml", "configtx.yaml", "shared-config.yaml"];
```

src/setup-docker/fabricXDockerWriter.ts
```diff
@@ -0,0 +1,38 @@
...
+      "party1-router.yaml",
+      "party1-batcher.yaml",
+      "party1-consenter.yaml",
+      "party1-assembler.yaml",
+      "committer-coordinator.yaml",
+      "committer-query-service.yaml",
+      "committer-sidecar-dev.yaml",
+      "committer-validator.yaml",
+      "committer-verifier.yaml",
+    ];
```
umegbewe: #799 adds `fxconfig.yaml` please add that to this list

umegbewe: I also noticed this names are stale, please update them

src/setup-docker/index.ts
```diff
@@ -70,6 +71,21 @@ export default class SetupDocker extends Command {
     this.log(`Fabric version is: ${global.fabricVersion}`);
     this.log(`Generating docker-compose network '${composeNetworkName}'...`);
 
+    // ======= Fabric-X =================================================================
+    if (global.provider === "fabric-x") {
+      this.log("Provider is fabric-x, delegating to FabricXDockerWriter...");
+      const fabricXWriter = new FabricXDockerWriter(this.templatesDir, this.outputDir, (msg) => this.log(msg));
+      await fabricXWriter.write(configExtended);
```
umegbewe, line 77: write() does not copy fabric-x-docker.sh, so the generated target has no command script for fablo.sh to dispatch to. Please copy `src/setup-docker/templates/fabric-x-docker.sh` to `fablo-target/fabric-x-docker.sh`

src/setup-docker/index.ts
```diff
@@ -70,6 +71,21 @@ export default class SetupDocker extends Command {
...
+
+      // ======= hooks ====================================================================
+      await this._copyHooks(configExtended.hooks);
+
+      this.log("Done & done !!! Try the network out: ");
+      this.log("-> cd fabric-x");
+      this.log("-> docker compose up -d");
```
umegbewe:
```suggestion
      this.log("-> fablo up - to start network");
      this.log("-> fablo help - to view all commands");
```

src/setup-docker/fabricXDockerWriter.ts
MaryamArif7: @HarK-github Since you used "shared-config.yml" in the merged code, I used the same. Let me know what you're planning to use, and I'll update it before my PR merges.

src/setup-docker/fabricXDockerWriter.ts
```diff
@@ -0,0 +1,52 @@
+import { FabloConfigExtended } from "../types/FabloConfigExtended";
+import { renderTemplate, getTemplatePath, getDestinationPath } from "../utils/templateUtils";
+
+export class FabricXDockerWriter {
+  constructor(private templatesDir: string, private outputDir: string, private log: (msg: string) => void) {}
+
+  public async write(configExtended: FabloConfigExtended): Promise<void> {
+    this.log("Generating Fabric-X network files...");
+
+    const staticFiles = [
```
umegbewe: on `staticFiles`, `staticConfigs` and `staticScripts`

Please don’t hard-code three separate file lists here. This makes the Fabric-X writer fragile, every new template file has to be added manually, and missing one will silently generate an incomplete target.

umegbewe: Use the same style as the rest of setup generation, render fabric-x-docker.sh, then recursively render everything under fabric-x/. The writer should own the Fabric-X target output.

umegbewe: i suggest

```sh
await this.renderTemplateFile("fabric-x-docker.sh", data);
await this.renderTemplateDirectory("fabric-x", data);
```

src/setup-docker/fabricXDockerWriter.ts
```diff
@@ -0,0 +1,52 @@
...
+      "query-service.yaml",
+      "sidecar.yaml",
+      "validator.yaml",
+      "verifier.yaml",
+    ];
+
+    for (const config of staticConfigs) {
+      const templatePath = getTemplatePath(this.templatesDir, `fabric-x/config/${config}`);
+      const destPath = getDestinationPath(this.outputDir, `fabric-x/config/${config}`);
+      await renderTemplate(templatePath, destPath, {});
```
umegbewe: Please pass the same template data to all Fabric-X templates instead of {}. Even if some files are static today, this blocks us from templating channel names, org names, ports, image tags, or paths later. It also makes the writer inconsistent for no benefit

umegbewe: use

```ts
const data = configExtended as unknown as Record<string, unknown>;
```

and pass data everywhere

src/setup-docker/index.ts
```diff
@@ -70,6 +71,26 @@ export default class SetupDocker extends Command {
...
     this.log(`Generating docker-compose network '${composeNetworkName}'...`);
 
+    // ======= Fabric-X =================================================================
+    if (global.provider === "fabric-x") {
+      this.log("Provider is fabric-x, delegating to FabricXDockerWriter...");
+      const fabricXWriter = new FabricXDockerWriter(this.templatesDir, this.outputDir, (msg) => this.log(msg));
+      await fabricXWriter.write(configExtended);
+      const fabricXDockerShTemplate = getTemplatePath(this.templatesDir, "fabric-x-docker.sh");
+      const fabricXDockerShDest = getDestinationPath(this.outputDir, "fabric-x-docker.sh");
+      await renderTemplate(fabricXDockerShTemplate, fabricXDockerShDest, {});
```
umegbewe: This belongs inside `FabricXDockerWriter`. `SetupDocker` should choose the provider and delegate, not know which Fabric-X template files need to be rendered. 

Otherwise Fabric-X generation is split across two classes and becomes harder to test.

umegbewe: i expect this to be closer to

```ts
if (global.provider === "fabric-x") {
  const fabricXWriter = new FabricXDockerWriter(this.templatesDir, this.outputDir, (msg) => this.log(msg));
  await fabricXWriter.write(configExtended);
  await this._copyHooks(configExtended.hooks);
  return;
}
```

src/setup-docker/index.ts
```diff
@@ -70,6 +71,26 @@ export default class SetupDocker extends Command {
...
+
+      // ======= hooks ====================================================================
+      await this._copyHooks(configExtended.hooks);
+
+      this.log("Done & done !!! Try the network out: ");
+      this.log("-> cd fabric-x");
+      this.log("-> docker compose up -d");
```
umegbewe: please remove this, misleading. A normal user should continue using fablo up, also the generated fabric-x/ directory is under fablo-target, not necessarily the user’s current directory.

src/setup-docker/index.ts
```diff
@@ -153,46 +180,49 @@ export default class SetupDocker extends Command {
     }
   }
 
+  async _createExplorerMaterial(
```
umegbewe: please revert this unrelated formatting change

src/setup-docker/index.ts
```diff
@@ -155,16 +170,6 @@ export default class SetupDocker extends Command {
...
-  try {
-    // Create explorer directory first
-    const explorerDir = getDestinationPath(this.outputDir, "fabric-config/explorer");
-    await fs.ensureDir(explorerDir);
-
-    if(!explorerDir){
-      this.log("Error: Explorer directory path is undefined");
-      return;
-    }
-
```
umegbewe, line 167: @HarK-github don't touch this, it's unrelated to this work

src/setup-docker/index.ts
```diff
@@ -188,11 +193,9 @@ export default class SetupDocker extends Command {
     if (!globalConfigPath) {
       this.error("Error: Global config path is undefined");
     }
+    await fs.ensureDir(path.dirname(globalConfigPath));
```
umegbewe: same unrelated comment, revert

src/setup-docker/fabricXDockerWriter.ts
```diff
@@ -0,0 +1,46 @@
...
+    const destPath = getDestinationPath(this.outputDir, file);
+    await renderTemplate(templatePath, destPath, data);
+  }
+
+  private async renderTemplateDirectory(dir: string, data: Record<string, unknown>): Promise<void> {
+    const templateDirPath = getTemplatePath(this.templatesDir, dir);
+    
+    if (!(await fs.pathExists(templateDirPath))) {
+      return;
+    }
```
umegbewe: don’t silently skip a missing Fabric-X template directory. If fabric-x/ is missing fail loudly. Returning here will produce an incomplete fablo-target and return success

Use `fs.readdir()` directly and bubble up the error

src/setup-docker/fabricXDockerWriter.ts
```diff
@@ -0,0 +1,46 @@
...
+    
+    const files = await fs.readdir(templateDirPath);
+    for (const file of files) {
+      const fullPath = path.join(templateDirPath, file);
+      const stat = await fs.stat(fullPath);
```
umegbewe: Use `fs.readdir(templateDirPath, { withFileTypes: true })` here. It avoids an extra stat call per file and matches the simpler recursive-directory style we want for this writer

umegbewe: something like

```ts
const entries = await fs.readdir(templateDirPath, { withFileTypes: true });

for (const entry of entries) {
  const relativePath = path.join(dir, entry.name);

  if (entry.isDirectory()) {
    await this.renderTemplateDirectory(relativePath, data);
  } else if (entry.isFile()) {
    await this.renderTemplateFile(relativePath, data);
  }
}
```

umegbewe (approved):

PR #808, opened 2026-08-12
Feat/jaiph docs parity

Efeguclu1:
## Summary

  Adds a Jaiph-based documentation parity workflow.

  - Runs deterministic documentation checks on relevant pull requests.
  - Uses Claude to review same-repository documentation PRs.
  - Runs a weekly repair pass that opens a separate documentation-sync PR.
  - Adds schema entries for supported CCaaS options.
  - Fixes CLI reference Markdown tables for Jekyll rendering.


.github/workflows/docs-parity.yml
```diff
@@ -0,0 +1,128 @@
...
+      - ".jaiph/**"
+  schedule:
+    # Mondays at 04:00 UTC.
+    - cron: "0 4 * * 1"
+  workflow_dispatch:
+    inputs:
+      mode:
+        description: "static | check | fix"
+        required: false
+        default: check
```
dzikowski: for now it should be only manual trigger so we can test it. We won't need the schedule I think, but we're going to need `pull_request` (so that part can stay as commented out)

.github/workflows/docs-parity.yml
```diff
@@ -0,0 +1,128 @@
...
+  # Deterministic parity checks. Runs everywhere, needs no secrets, so fork
+  # PRs get the same protection as branch PRs.
+  static-checks:
+    if: github.event_name != 'schedule'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up Jaiph
+        run: npm install --global jaiph@0.11.0
```
dzikowski: you can use the dedicated action:

```
  - uses: jaiphlang/jaiph/actions/setup-jaiph@v0.13.0
    with:
      version: 0.13.0        # semver, a release tag (v0.13.0), or 'nightly'
```

.github/workflows/docs-parity.yml
```diff
@@ -0,0 +1,128 @@
...
+
+      - name: Validate workflow definitions
+        run: |
+          jaiph compile .jaiph
+          jaiph test .jaiph/docs_parity.test.jh
+
+      - name: Run static documentation parity checks
+        run: jaiph run --unsafe .jaiph/docs_parity.jh static
```
dzikowski:
```suggestion
        run: .jaiph/docs_parity.jh --unsafe static
```

.github/workflows/docs-parity.yml
dzikowski: and keep it maybe as a single job to keep things as simple as possible. Allow the agent to update files, then at the end as a step you can run `git diff` so we see changes recommended by the agent

dzikowski (approved):

PR #811, opened 2026-08-19
feat(fabric-x): Add fablo init fabric-x

### LINKED ISSUE #809: feat(fabric-x): Add `fablo init fabric-x`
Let users generate the minimal Fabric-X starter config with `fablo init fabric-x`, using the same init style as `fablo init node`.

- Add `fabric-x` to the valid init options in `src/commands/init/index.ts`.
- Add a `fabricX` boolean to the init option parsing result.
- Update `getDefaultFabloConfig` so it can return a hardcoded Fabric-X starter topology when the Fabric-X init option is selected.
- Keep the generated output as a normal `fablo-config.json` that points at the same Fablo schema.
- Set `global.provider` to `fabric-x` in the generated config.
- Set hardcoded PoC values for Fabric-X:
  - `global.fabricVersion: "3.1.0"`
  - `global.tls: true`
  - `global.engine: "docker"`
  - `global.peerDevMode: false`
  - one application org named `Org1` with `Org1MSP` and `org1.example.com`
  - one channel named `mychannel`
  - empty `chaincodes`
  - empty `hooks`

MaryamArif7:
fixes  #809

This adds a new init option that generates a minimal starter config for Fabric X, following the same pattern as the existing init node option.

Changes:

1)Added fabric-x as a valid option in the init command argument parser, alongside the existing ones

2)Added a fabricX boolean to the parsed flags result.

3)Split the config generation into two functions, one for the regular Fabric starter config and one for the new Fabric X starter config, and getDefaultFabloConfig now picks between them based on the fabricX flag.

4)fixed values for the proof of concept: fabricVersion 3.1.0, tls true, engine docker, peerDevMode false. It includes one application org called Org1 with Org1MSP and org1.example.com, one channel called mychannel, and empty chaincodes and hooks. The output is still a normal fablo-config.json file pointing at the same schema, so it validates the same way as any other config.

### DIFF
--- a/src/commands/init/index.ts
+++ b/src/commands/init/index.ts
@@ -95,7 +95,6 @@ function getDefaultFabricXConfig(): FabloConfigJson {
         orderers: [
           {
             groupName: "group1",
-
             type: "BFT",
             instances: 1,
             prefix: "orderer",
--- a/src/commands/init/index.ts
+++ b/src/commands/init/index.ts
@@ -8,7 +8,7 @@ import * as fs from "fs-extra";
 import { version } from "../../../package.json";
 import { schema } from "../../config";
 
-function getDefaultFabloConfig(): FabloConfigJson {
+function getFabricConfig(): FabloConfigJson {
   return {
     $schema: `https://github.com/hyperledger-labs/fablo/releases/download/${version}/schema.json`,
     global: {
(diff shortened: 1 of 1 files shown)


MaryamArif7:
For the fabric-x starter config, OrgJson.ca is required by the schema, so I added a placeholder { prefix: "ca", db: "sqlite" } to satisfy it

MaryamArif7:
@umegbewe 
for the testing :

i ran these locally fablo init fabric-x generates the expected config:

```
mkdir -p /tmp/fabricx-test && cd /tmp/fabricx-test
node ~/fablo/bin/run.mjs init fabric-x
cat fablo-config.json
```

image

config generated :

image

image

MaryamArif7:
@umegbewe not related to this issue but i keep hitting  error for root-owned files under crypto and data, docker creates them before my script can. Thinking of cleaning up with a root container instead of plain rm -rf. Good idea or is there a better fix? 
 (even tho it was cleaned up before with down)
```
maryam@MARY7:~/fablo/src/setup-docker/templates$ ./fabric-x-docker.sh up
Removing incomplete Fabric-X crypto material...
rm: cannot remove '/home/maryam/fablo/src/setup-docker/templates/fabric-x/crypto/ordererOrganizations/orderer.example.com/msp/tlscacerts/tlsca.orderer.example.com-cert.pem':
 Permission denied
```
if yes then i can add in this pr

src/commands/init/index.ts
```diff
@@ -70,6 +70,51 @@ function getDefaultFabloConfig(): FabloConfigJson {
...
+    ],
+    chaincodes: [],
+    hooks: {},
+  };
+}
+
+function getDefaultFabloConfig(fabricX = false): FabloConfigJson {
+  return fabricX ? getFabricXConfig() : getFabricConfig();
+}
+
```
umegbewe: This boolean parameter makes the default config API less clear. Please keep classic Fabric as the no-argument default and choose the Fabric-X config explicitly in copySampleConfig.

like this

```
function getDefaultFabloConfig(): FabloConfigJson {
  return getFabricConfig();
}

function getDefaultFabricXConfig(): FabloConfigJson {
  return { ... };
}
```

umegbewe: then

```ts
let fabloConfigJson = flags.fabricX ? getDefaultFabricXConfig() : getDefaultFabloConfig();
```

src/commands/init/index.ts
```diff
@@ -70,6 +70,51 @@ function getDefaultFabloConfig(): FabloConfigJson {
   };
 }
 
+
+function getFabricXConfig(): FabloConfigJson {
```
umegbewe: This config doesn't contain orderer org/orderer group for compatibility so will likely fail. 

Either relax the validation to not check that for provider == fabric-x or add it here for compatibility

umegbewe (approved):

PR #816, opened 2026-08-27
 test: add Fabric-X e2e test script and CI integration

### LINKED ISSUE #812: feat (fabric-x):  Add Fabric-X e2e regression coverage
Add a Fabric-X network test script under e2e-network/docker, for example `e2e-network/docker/test-08-fabric-x.sh` and wire it into the current test on push CI.

It should follow the structure of the existing network tests

The scenario should exercise Fabric-X the way a user does. From the temporary directory, run `fablo.sh init fabric-x`, `fablo.sh validate`, and `fablo.sh up`. After startup, assert that the generated Fabric-X artifacts are present and non-empty, at minimum `config-block.pb.bin`, `shared_config.binpb`, and `client-tls-ca.pem`, and verify the generated Fabric-X containers are healthy.

Then cover the namespace and lifecycle behavior: `namespace list` should work before namespace creation and show zero namespaces, `namespace` init should commit the default namespace, a later `namespace list` should show mynamespace, rerunning `namespace init` should be idempotent, rerunning `up` should skip artifact generation, `stop` and `start` should preserve a usable network, and `reset` should clear namespace state.

HarK-github:
Closes #812 
This PR adds a new E2E test suite (`test-08-fabric-x`) for the Fabric-X network profile and integrates it into the CI pipeline.

### 1. Test Harness

- Created `e2e-network/docker/test-08-fabric-x.sh` with isolated workspace, `set -euo pipefail`, trap-based cleanup, artifact assertions, container health checks, and full `namespace` lifecycle regression suite (zero state → init → idempotency → cached up → stop/start → reset).

### 2. CI Integration

- Added `test-08-fabric-x` job to `.github/workflows/test-on-push.yml` with artifact upload of logs and tmpdir on failure. Removed redundant `mkdir`/`chmod` logic from `base-functions.sh` and added `max_connections=500` to Postgres for connection-pool exhaustion.

### 3.  UID/GID Resolution Failure

- **Problem:** Orderer containers failed with `unhealthy` due to LevelDB permission panics. The existing `user: "${UID:-1000}:${GID:-1000}"` didn't work because bash's `UID`/`GID` are shell built-ins, not exported environment variables. Compose fell back to `1000:1000` while CI runners use UID `1001`, causing a mismatch on bind-mounted directories.
- **Solution:** Replaced with `user: "${FABRIC_X_UID:-1000}:${FABRIC_X_GID:-1000}"` across all orderer services and `committer-org1-sidecar`. These variables are exported in `base-functions.sh`, ensuring container UID matches host UID.

**Status**
- [x] All container health and lifecycle assertions are passing locally and in the CI run.
- [x] Workflow integrates cleanly with other test jobs.

### DIFF
--- a/e2e-network/docker/test-08-fabric-x.sh
+++ b/e2e-network/docker/test-08-fabric-x.sh
@@ -1,0 +1,128 @@
+#!/usr/bin/env bash
+
+set -e
+
+# Test Harness & Workspace Scaffolding
+
+TEST_TMP="$(rm -rf "$0.tmpdir" && mkdir -p "$0.tmpdir" && (cd "$0.tmpdir" && pwd))"
+TEST_LOGS="$(mkdir -p "$0.logs" && (cd "$0.logs" && pwd))"
+FABLO_HOME="$TEST_TMP/../../.."
+
+export FABLO_HOME
+
+dumpLogs() {
+  echo "Saving logs of $1 to $TEST_LOGS/$1.log"
+  mkdir -p "$TEST_LOGS"
+  docker logs "$1" >"$TEST_LOGS/$1.log" 2>&1
+}
+
+networkDown() {
+  (for name in $(docker ps --format '{{.Names}}'); do dumpLogs "$name"; done)
+  (cd "$TEST_TMP" && "$FABLO_HOME/fablo.sh" down || true)
+}
+
+trap networkDown EXIT
+trap 'networkDown ; echo "Test failed" ; exit 1' ERR SIGINT
+
+echo "Starting Fabric-X Network Initialization..."
+"$FABLO_HOME/fablo-build.sh"
+
+# Scaffold the network profile
+(cd "$TEST_TMP" && "$FABLO_HOME/fablo.sh" init fabric-x)
+
+# Validate the generated profile
+(cd "$TEST_TMP" && "$FABLO_HOME/fablo.sh" validate)
+
+# Generate artifacts and spin up Docker containers
+(cd "$TEST_TMP" && "$FABLO_HOME/fablo.sh" up)
+
+echo "Network started successfully."
+
+# Artifact Verification & Container Health Checks
+echo "Running Artifact Verification..."
+
+assert_non_empty() {
+  local file="$1"
+  if [ -z "$file" ] || [ ! -s "$file" ]; then
+    echo "Error: Artifact missing or empty."
+    exit 1
+  fi
+  echo "Verified artifact exists and is non-empty: $file"
+}
+
+# Locate generated Fabric-X artifacts
+CONFIG_BLOCK=$(find "$TEST_TMP/fablo-target" -name "config-block.pb.bin" | head -n 1)
+SHARED_CONFIG=$(find "$TEST_TMP/fablo-target" -name "shared_config.binpb" | head -n 1)
+CLIENT_TLS=$(find "$TEST_TMP/fablo-target" -name "client-tls-ca.pem" | head -n 1)
+
+assert_non_empty "$CONFIG_BLOCK"
+assert_non_empty "$SHARED_CONFIG"
+assert_non_empty "$CLIENT_TLS"
+
+echo "Running Container Health Checks..."
+# Check that containers for this project are running
+# Typically fablo networks use a common label or prefix. Here we verify that at least some orderer/peer components are up.
+RUNNING_CONTAINERS=$(docker ps --format '{{.Names}}')
+
+if ! echo "$RUNNING_CONTAINERS" | grep -q "orderer"; then
+  echo "Error: No orderer containers found running."
+  exit 1
+fi
+
+if ! echo "$RUNNING_CONTAINERS" | grep -q "peer\|committer"; then
+  echo "Error: No peer/committer containers found running."
+  exit 1
+fi
+
+echo "All required Fabric-X containers are running."
+
+# Namespace & Lifecycle Regression Validation
+echo "Running Namespace & Lifecycle Regression Validation..."
+
+run_fablo() {
+  (cd "$TEST_TMP" && "$FABLO_HOME/fablo.sh" "$@")
+}
+
+echo "1. Zero State"
+NS_LIST=$(run_fablo namespace list 2>&1 || true)
+if echo "$NS_LIST" | grep -q "mynamespace"; then
+  echo "Error: Expected 0 namespaces in zero state. Found: $NS_LIST"
+  exit 1
+fi
+
+echo "2. Commit State"
+run_fablo namespace init
+
+echo "3. Post-Init Query"
+NS_LIST=$(run_fablo namespace list 2>&1)
+if ! echo "$NS_LIST" | grep -q "mynamespace"; then
+  echo "Error: mynamespace not found after init. Output: $NS_LIST"
+  exit 1
+fi
+
+echo "4. Idempotency"
+run_fablo namespace init
+
+echo "5. Cache / Up Skip"
+run_fablo up
+
+echo "6. Stop / Start"
+run_fablo stop
+run_fablo start
+NS_LIST=$(run_fablo namespace list 2>&1)
+if ! echo "$NS_LIST" | grep -q "mynamespace"; then
+  echo "Error: mynamespace not found after stop/start. Output: $NS_LIST"
+  exit 1
+fi
+
+echo "7. Reset State"
+run_fablo reset
+# After reset, it might error if the target is deleted, or return an empty list.
+# We just want to ensure it doesn't show 'mynamespace' anymore.
+NS_LIST_RESET=$(run_fablo namespace list 2>&1 || true)
+if echo "$NS_LIST_RESET" | grep -q "mynamespace"; then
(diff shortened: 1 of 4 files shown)


e2e-network/docker/test-08-fabric-x.sh
```diff
@@ -0,0 +1,181 @@
...
+
+# Namespace & Lifecycle Regression Validation
+echo "Running Namespace & Lifecycle Regression Validation..."
+
+echo "1. Zero State"
+NS_LIST=$(run_fablo namespace list 2>&1 || true)
+if echo "$NS_LIST" | grep -q "mynamespace"; then
+  echo "Error: Expected 0 namespaces in zero state. Found: $NS_LIST"
+  exit 1
+fi
```
umegbewe: This can pass even if fablo.sh namespace list is broken. Tis should prove that namespace list succeeds before init and after reset. Please remove || true here and assert Installed namespaces (0 total). Same concern applies to the reset check around line 588.

e2e-network/docker/test-08-fabric-x.sh
```diff
@@ -0,0 +1,181 @@
...
+    sleep 10
+  else
+    echo "fablo up failed (attempt $i). Final attempt failed."
+  fi
+done
+
+if [ "$UP_SUCCESS" = false ]; then
+  echo "Error: fablo up failed after 5 attempts."
+  exit 1
+fi
```
umegbewe: What's the reason for retrying here?

src/setup-docker/templates/fabric-x/scripts/base-functions.sh
```diff
@@ -61,22 +61,32 @@ generateArtifacts() {
     configtxgen --channelID mychannel --profile OrgsChannel \
     --outputBlock /config/crypto/config-block.pb.bin \
     --configPath /config
+
+  echo "Fixing permissions for crypto material to ensure containers can read TLS keys..."
+  chmod -R a+rX "$FABRIC_X_ROOT/crypto" || true
```
umegbewe: I don't feel good about this, If Postgres needs a different user/volume model, split that into a separate runtime fix PR with a clear reason.

src/setup-docker/templates/fabric-x/docker-compose.yaml
```diff
@@ -128,29 +131,28 @@ services:
...
-      retries: 10
+      retries: 30
     networks:
       - fabric-x
 
   committer-org1-verifier:
     image: ghcr.io/hyperledger/fabric-x-committer:1.0.3
     pull_policy: missing
     restart: unless-stopped
-    user: "${UID:-1000}:${GID:-1000}"
```
umegbewe, line 153: what's the reason for this change?

e2e-network/docker/test-08-fabric-x.sh
```diff
@@ -0,0 +1,181 @@
...
+waitForContainer "orderer-consenter" "Consensus network service is starting"
+waitForContainer "orderer-assembler" "Assembler network service is starting"
+
+# Wait for committer stack
+waitForContainer "fabric-x-committer-org1-db-1" "database system is ready to accept connections"
+waitForContainer "fabric-x-committer-org1-verifier-1" "Serving gRPC on"
+waitForContainer "fabric-x-committer-org1-validator-1" "Serving gRPC on"
+waitForContainer "fabric-x-committer-org1-coordinator-1" "Serving gRPC on"
+waitForContainer "fabric-x-committer-org1-query-service-1" "Serving gRPC on"
+waitForContainer "fabric-x-committer-org1-sidecar-1" "Updated dynamic TLS"
```
umegbewe: `docker compose up -d --wait` already depends on health checks. These log strings are image-version-specific and may fail even when the services are healthy. 

Either assert Docker health status for the known containers or rely on namespace list or init as the functional readiness check.

e2e-network/docker/test-08-fabric-x.sh
```diff
@@ -0,0 +1,181 @@
...
+assert_non_empty() {
+  local file="$1"
+  if [ -z "$file" ] || [ ! -s "$file" ]; then
+    echo "Error: Artifact missing or empty: $file"
+    exit 1
+  fi
+  echo "Verified artifact exists and is non-empty: $file"
+}
+
+CONFIG_BLOCK=$(find "$TEST_TMP/fablo-target" -name "config-block.pb.bin" | head -n 1 || true)
```
umegbewe: Use` "$TEST_TMP/fablo-target/fabric-x/crypto/config-block.pb.bin"`etc. 

The test should catch layout regressions, not find the first matching file somewhere under the target.

e2e-network/docker/test-08-fabric-x.sh
```diff
@@ -0,0 +1,181 @@
...
+  exit $exit_code
+}
+
+trap networkDown EXIT SIGINT
+
+echo "Starting Fabric-X Network Initialization..."
+"$FABLO_HOME/fablo-build.sh"
+
+(cd "$TEST_TMP" && "$FABLO_HOME/fablo.sh" init fabric-x)
+(cd "$TEST_TMP" && "$FABLO_HOME/fablo.sh" validate)
```
umegbewe: This should follow the existing e2e structure more closely: define a networkUp() function and call it after the traps are installed. Most existing tests keep setup, startup, cleanup, log dumping, and assertions in predictable helper blocks. That makes failures easier to compare across e2e tests.

e2e-network/docker/test-08-fabric-x.sh
```diff
@@ -0,0 +1,181 @@
...
+
+# Helper function to run Fablo inside temp directory
+run_fablo() {
+  (cd "$TEST_TMP" && "$FABLO_HOME/fablo.sh" "$@")
+}
```
umegbewe, line 18: This helper is useful, but it should be introduced before the first fablo.sh call and used consistently for init, validate, up, down, and namespace commands. Mixing raw subshell calls with run_fablo makes the script less consistent than the existing e2e tests.

e2e-network/docker/test-08-fabric-x.sh
```diff
@@ -0,0 +1,181 @@
...
+echo "3. Post-Init Query"
+NS_LIST=$(run_fablo namespace list 2>&1 || true)
+if ! echo "$NS_LIST" | grep -q "mynamespace"; then
+  echo "Error: mynamespace not found after init. Output: $NS_LIST"
+  exit 1
+fi
+
+echo "4. Idempotency"
+run_fablo namespace init
+NS_LIST_IDEMPOTENT=$(run_fablo namespace list 2>&1 || true)
```
umegbewe: I would tone this down to match the existing e2e style. 

The current scripts generally test behavior directly with concise comments/helper names rather than numbered scenario. 

This is a nit, but the test should feel like the rest of e2e-network/docker.

umegbewe: also instead of repeating inline grep assertions, add a small assertion helper or reuse the style from existing e2e helpers such as expectCommand. 

The important part is that command success and expected output should be checked together, without || true hiding broken commands.

e2e-network/docker/test-08-fabric-x.sh
```diff
@@ -0,0 +1,181 @@
...
+  set +e
+  echo "Executing teardown. Capturing container logs..."
+  docker ps -a --filter "label=com.docker.compose.project=fabric-x" --format '{{.Names}}' | while read -r name; do
+    [ -n "$name" ] && dumpLogs "$name"
+  done
+  ( cd "$TEST_TMP" && "$FABLO_HOME/fablo.sh" down ) || true
+  if [ $exit_code -ne 0 ]; then
+    echo "❌ Test failed with exit code $exit_code"
+  fi
+  exit $exit_code
```
umegbewe: Cleanup functions in the existing e2e tests generally clean up and let the trap handler control the final exit path. 

Having `networkDown()` call exit directly makes the function harder to reuse safely. Prefer a cleanup function that only cleans up, then preserve the exit code in the trap wrapper.

src/setup-docker/templates/fabric-x/scripts/base-functions.sh
HarK-github: this was a temporary workaround, removed it.

e2e-network/docker/test-08-fabric-x.sh
HarK-github: replaced with health checks.

src/setup-docker/templates/fabric-x/docker-compose.yaml
HarK-github: Currently UID fallback fails in ci because the runner operates on UID 1001, which causes permission panics on host bind-mounts. The official postgres image crashes when forced to run a host UID, as its entrypoint script must boot as root to run internal initialization steps before dropping privileges to postres's UID.
To resolve this,  i  removed the forced uid from the postgres container in favour of  docker named volume ,which  resolved the host permission conflicts.

HarK-github: Attaching the error logs: 
image

umegbewe (approved):

PR #824, opened 2026-09-07
feat: Making the Fabric-X channel configurable

### LINKED ISSUE #817: feat (fabric-x): Make the Fabric-X channel configurable
Fabric-X generation currently assumes the default channel. Research where the channel name appears in the generated Fabric-X files and update the generator so the value comes from the normal Fablo config.

The implementation should keep the default generated network working, but a user should also be able to change the channel name in `fablo-config.json` and still run the Fabric-X flow successfully.

MaryamArif7:
fixes #817

**What it does**

Fabric-X generation was hardcoding the channel name (`mychannel`) instead of reading it from `fablo-config.json`. This makes the channel name configurable.

 **Changes**

1. `src/setup-docker/templates/fabric-x/scripts/base-functions.sh` — `configtxgen --channelID` now uses `<%= channels[0].name %>` instead of the hardcoded `mychannel`.
2. `src/setup-docker/templates/fabric-x/fxconfig.yaml` — `channel:` field now uses `<%= channels[0].name %>` instead of the hardcoded `mychannel`.

### DIFF
--- a/src/setup-docker/templates/fabric-x/fxconfig.yaml
+++ b/src/setup-docker/templates/fabric-x/fxconfig.yaml
@@ -7,7 +7,7 @@ msp:
 
 orderer:
   address: orderer-router:7050
-  channel: mychannel
+  channel: <%= channels[0].name %>
   tls:
     enabled: true
     clientCert: /tls/server.crt
--- a/src/setup-docker/templates/fabric-x/scripts/base-functions.sh
+++ b/src/setup-docker/templates/fabric-x/scripts/base-functions.sh
@@ -58,7 +58,7 @@ generateArtifacts() {
   docker run --rm --user "$(id -u):$(id -g)" \
     -v "$FABRIC_X_ROOT:/config" \
     "$TOOLS_IMAGE" \
-    configtxgen --channelID <%= channels[0].name %> --profile <%= channels[0].profileName %> \\
+    configtxgen --channelID <%= channels[0].name %> --profile <%= channels[0].profileName %> \
     --outputBlock /config/crypto/config-block.pb.bin \
     --configPath /config
 }
--- a/src/setup-docker/templates/fabric-x/scripts/base-functions.sh
+++ b/src/setup-docker/templates/fabric-x/scripts/base-functions.sh
@@ -58,7 +58,7 @@ generateArtifacts() {
   docker run --rm --user "$(id -u):$(id -g)" \
     -v "$FABRIC_X_ROOT:/config" \
     "$TOOLS_IMAGE" \
-    configtxgen --channelID mychannel --profile OrgsChannel \
+    configtxgen --channelID <%= channels[0].name %> --profile <%= channels[0].profileName %> \\
     --outputBlock /config/crypto/config-block.pb.bin \
     --configPath /config
 }
--- a/src/setup-docker/templates/fabric-x/configtx.yaml
+++ b/src/setup-docker/templates/fabric-x/configtx.yaml
@@ -149,7 +149,7 @@ Channel: &ChannelDefaults
     <<: *ChannelCapabilities
 
 Profiles:
-  OrgsChannel:
+  <%= channels[0].profileName %>:
     <<: *ChannelDefaults
     Orderer:
       <<: *OrdererDefaults


MaryamArif7:
**Research: where the channel name (`mychannel`) appears**
1. Hardcoded in source templates (`src/setup-docker/templates/fabric-x*`):

`fxconfig.yaml:10           channel: mychannel`
`scripts/base-functions.sh:61   configtxgen --channelID mychannel`

 2.In the generated target for the fabric-x:

1. In the `fablo/fablo-target/fabric-x/fxconfig.yaml` file

Image

`2.fabric-x/crypto/config-block.pb.bin`
`3.fabric-x/data/orderers/party1-consenter/chains/consensus/blockfile_000000`
`4.fabric-x/data/orderers/party1-router/config.pb.bin/block-0.config.pb.bin`
`5.fabric-x/data/orderers/party1-batcher/config.pb.bin/block-0.config.pb.bin`
`6.fabric-x/data/orderers/party1-batcher/wal/0000000000000001.wal`
`7.fabric-x/data/committer-org1/sidecar-ledger/chains/fabric-x-committer/blockfile_000000`

MaryamArif7:
few notes :
"channelID <%= channels[0].name %> --profile <%= channels[0].profileName %> "
->for the profileName i have followed the same pattern of fabric-classic
->channels[0] ->will get 1st channel only ,because fabric-x is validated elsewhere to allow exactly one channel

MaryamArif7:
**Few notes :**
"channelID <%= channels[0].name %> --profile <%= channels[0].profileName %> "
->for the profileName i have followed the same pattern of fabric-classic
->channels[0] ->will get 1st channel only ,because fabric-x is validated elsewhere to allow exactly one channel

## Verification

**1) Default channel , `mychannel`**

- Rebuilt the local image
- Pruned the existing network
- Regenerated the default config using `./fablo.sh init fabric-x`
- Confirmed the default channel name is set correctly

image

- Confirmed the generated files pick up the default channel
- Ran `up` and all namespace commands successfully

**2) Custom channel name  `mychanneltest`**

- Changed the channel name in the default init config, then also confirmed it can be changed directly in `fablo-config.json` after generation
- Rebuilt the local image
- Confirmed the new channel name is reflected in the generated files (examples below)

image
image

- Ran `up` and namespace commands ,everything works

src/setup-docker/templates/fabric-x/fxconfig.yaml
```diff
@@ -7,7 +7,7 @@ msp:
 
 orderer:
   address: orderer-router:7050
-  channel: mychannel
+  channel: <%= channels[0].name %>
```
umegbewe: Don't reach, channels[0] directly from the template. Add a small fabric-x template model in  FabricXDockerWriter

so the template should consume something like

```yaml
channel: <%- fabricX.channelName %>
```

src/setup-docker/templates/fabric-x/scripts/base-functions.sh
```diff
@@ -58,7 +58,7 @@ generateArtifacts() {
   docker run --rm --user "$(id -u):$(id -g)" \
     -v "$FABRIC_X_ROOT:/config" \
     "$TOOLS_IMAGE" \
-    configtxgen --channelID mychannel --profile OrgsChannel \
+    configtxgen --channelID <%= channels[0].name %> --profile <%= channels[0].profileName %> \
```
umegbewe: Same issue here. The channel value should come from the Fabric-X model, not raw channels[0].

umegbewe:
Looks okay, but i'd expect this PR to be modified into something like this in `FabricXDockerWriter`:

```
interface FabricXTemplateModel {
  channelName: string;
  channelProfileName: string;
}
```

and then render templates from:

```
<%- fabricX.channelName %>
<%- fabricX.channelProfileName %>
```

not from channels[0] and some tests

src/types/FabloConfigExtended.ts
```diff
@@ -209,3 +209,7 @@ export interface FabloConfigExtended {
   chaincodes: ChaincodeConfig[];
   hooks: HooksConfig;
 }
+export interface FabricXTemplateModel {
+  channelName: string;
+  channelProfileName: string;
+}
```
umegbewe: keep this local to fabricXDockerWriter.ts

src/setup-docker/fabricXDockerWriter.ts
```diff
@@ -1,15 +1,26 @@
...
 
+
+export const getFabricXTemplateModel = (configExtended: FabloConfigExtended): FabricXTemplateModel => {
+  const [channel] = configExtended.channels;
+ 
+  return {
+    channelName: channel.name,
+    channelProfileName: channel.profileName,
+  };
+};
```
umegbewe, line 20: This assumes channels[0] exists and will crash with a generic Cannot read properties of undefined if generation is called with a bad config

add a defensive check like this

```ts
if (!channel) {
  throw new Error("Fabric-X generation requires exactly one channel.");
}
```

src/setup-docker/templates/fabric-x/scripts/base-functions.sh
```diff
@@ -58,7 +58,7 @@ generateArtifacts() {
   docker run --rm --user "$(id -u):$(id -g)" \
     -v "$FABRIC_X_ROOT:/config" \
     "$TOOLS_IMAGE" \
-    configtxgen --channelID mychannel --profile OrgsChannel \
+    configtxgen --channelID <%- fabricX.channelName %>  --profile <%- fabricX.channelProfileName %> \
```
umegbewe: looks like there's a space or two before --profile

src/setup-docker/fabricXDockerWriter.ts
```diff
@@ -3,13 +3,29 @@ import { renderTemplate, getTemplatePath, getDestinationPath } from "../utils/te
 import * as fs from "fs-extra";
 import * as path from "path";
 
+export interface FabricXTemplateModel {
+  channelName: string;
+  channelProfileName: string;
+}
```
umegbewe, line 9:
```suggestion
}

```

umegbewe (approved):

PR #828, opened 2026-09-09
feat:Add config-driven Fabric-X namespaces

### LINKED ISSUE #819: feat (fabric-x):  Add config-driven Fabric-X namespaces
Namespaces should be defined from the Fablo config rather than being hardcoded in the generated scripts. Research a clean way to represent Fabric-X namespaces in the shared schema without creating a separate Fabric-X schema.

Starting point: add a top-level `namespaces` section to `fablo-config.json`

The result should let users declare at least one namespace in `fablo-config.json`, generate the needed Fabric-X configuration from it, and create that namespace through the Fabric-X namespace command flow

MaryamArif7:
fixes #819 

This PR adds support for configuring Fabric-X namespaces through the Fablo configuration.

Changes

1. Added support for defining namespaces in the config.
2. Added validation for namespace configuration.
3. Added namespace initialization during network setup.
4. Supports multiple namespaces.

This makes it easier to set up Fabric-X namespaces without needing to configure them manually.

Changed files

1. schema.json: added namespaces, next to channels and chaincodes, one shared schema
2. FabloConfigJson.ts / FabloConfigExtended.ts: added namespace types
3. defaults.ts: added the org-list-to-policy function
4. extendConfig.ts: calls namespace resolution for fabric-x only, using only the channel's orgs
5. validate/index.ts: the fabric-x namespace checks above
6. init/index.ts: default namespace now uses orgs
7. base-functions.sh / base-help.sh: generated script and help text reflect config-driven namespaces
8. fabric-x-docker.sh: passes an optional name through to namespace init
9. test-08-fabric-x.sh: added targeted init and unknown-name checks to the real Docker e2e test

### DIFF
--- a/src/commands/init/index.ts
+++ b/src/commands/init/index.ts
@@ -113,9 +113,15 @@ function getDefaultFabricXConfig(): FabloConfigJson {
         ],
       },
     ],
+    namespaces: [
+      {
+        name: "mynamespace",
+      },
+
+    ],
     chaincodes: [],
     hooks: {},
-  };
+  }
 }
 
 function getDefaultFabloConfig(): FabloConfigJson {
--- a/src/commands/validate/index.ts
+++ b/src/commands/validate/index.ts
@@ -583,6 +583,24 @@ export default class Validate extends Command {
         message: `fabric-x requires exactly one channel found ${channels.length}.`,
       });
     }
+        const namespaces = networkConfig.namespaces ?? [];
+    if (namespaces.length < 1) {
+      this.emit(validationErrorType.ERROR, {
+        category: validationCategories.GENERAL,
+        message: "fabric-x requires at least one entry in 'namespaces'.",
+      });
+    }
+ 
+    const namespaceNames = new Set<string>();
+    namespaces.forEach((namespace) => {
+      if (namespaceNames.has(namespace.name)) {
+        this.emit(validationErrorType.ERROR, {
+          category: validationCategories.GENERAL,
+          message: `Duplicate namespace '${namespace.name}' found. Namespace names must be unique.`,
+        });
+      }
+      namespaceNames.add(namespace.name);
+    });
   }
   _validateExplorer(global: GlobalJson, orgs: OrgJson[]): void {
     if (global.tools?.explorer === true) {
--- a/src/extend-config/defaults.ts
+++ b/src/extend-config/defaults.ts
@@ -34,4 +34,9 @@ export default {
       return capabilities.isV2 ? undefined : `AND (${orgs.map((o) => `'${o.mspName}.member'`).join(", ")})`;
     },
   },
+   namespace: {
+    policy(orgs: OrgConfig[]): string {
+      return `AND(${orgs.map((o) => `'${o.mspName}.member'`).join(",")})`;
+    },
+  },
 };
--- a/src/extend-config/extendConfig.ts
+++ b/src/extend-config/extendConfig.ts
@@ -6,7 +6,7 @@ import extendChannelsConfig from "./extendChannelsConfig";
 import extendChaincodesConfig, { checkUniqueChaincodeNames } from "./extendChaincodesConfig";
 import extendHooksConfig from "./extendHooksConfig";
 import { distinctOrdererHeads, mergeOrdererGroups } from "./mergeOrdererGroups";
-
+import extendNamespacesConfig from "./extendNamespacesConfig ";
 const extendConfig = (json: FabloConfigJson): FabloConfigExtended => {
   const {
     global: globalJson,
@@ -14,6 +14,7 @@ const extendConfig = (json: FabloConfigJson): FabloConfigExtended => {
     channels: channelsJson,
     chaincodes: chaincodesJson,
     hooks: hooksJson,
+    namespaces: namespacesJson,
   } = json;
 
   const global = extendGlobal(globalJson);
@@ -25,7 +26,7 @@ const extendConfig = (json: FabloConfigJson): FabloConfigExtended => {
   checkUniqueChaincodeNames(chaincodesJson);
   const chaincodes = extendChaincodesConfig(chaincodesJson, channels, global);
   const hooks = extendHooksConfig(hooksJson);
-
+  const namespaces = extendNamespacesConfig(namespacesJson ?? [], orgs);
   return {
     global,
     ordererGroups,
@@ -34,6 +35,7 @@ const extendConfig = (json: FabloConfigJson): FabloConfigExtended => {
     channels,
     chaincodes,
     hooks,
+    namespaces,
   };
 };
 
--- a/src/types/FabloConfigExtended.ts
+++ b/src/types/FabloConfigExtended.ts
@@ -199,7 +199,10 @@ export interface HooksConfig {
   postGenerate: string;
   postStart: string;
 }
-
+export interface NamespaceConfig {
+  name: string;
+  policy: string;
+}
 export interface FabloConfigExtended {
   global: Global;
   ordererGroups: OrdererGroup[];
@@ -208,4 +211,5 @@ export interface FabloConfigExtended {
   channels: ChannelConfig[];
   chaincodes: ChaincodeConfig[];
   hooks: HooksConfig;
+  namespaces: NamespaceConfig[]
 }
--- a/src/extend-config/extendNamespacesConfig .ts
+++ b/src/extend-config/extendNamespacesConfig .ts
@@ -1,0 +1,26 @@
+import { NamespaceJson } from "../types/FabloConfigJson";
+import { NamespaceConfig, OrgConfig } from "../types/FabloConfigExtended";
+import defaults from "./defaults";
+
+const extendNamespaceConfig = (namespaceJsonFormat: NamespaceJson, orgsTransformed: OrgConfig[]): NamespaceConfig => ({
+  name: namespaceJsonFormat.name,
+  policy: namespaceJsonFormat.policy ?? defaults.namespace.policy(orgsTransformed),
+});
+
+export const checkUniqueNamespaceNames = (namespacesJsonFormat: NamespaceJson[]): void => {
+  const namespaceNames = new Set<string>();
+
+  namespacesJsonFormat.forEach((namespace) => {
+    if (namespaceNames.has(namespace.name)) {
+      throw new Error(`Duplicate namespace '${namespace.name}' found. Namespace names must be unique.`);
+    }
+    namespaceNames.add(namespace.name);
+  });
+};
+
+const extendNamespacesConfig = (namespacesJsonFormat: NamespaceJson[], orgsTransformed: OrgConfig[]): NamespaceConfig[] => {
+  checkUniqueNamespaceNames(namespacesJsonFormat);
+  return namespacesJsonFormat.map((ns) => extendNamespaceConfig(ns, orgsTransformed));
+};
+
+export default extendNamespacesConfig;
--- a/src/setup-docker/fabricXDockerWriter.ts
+++ b/src/setup-docker/fabricXDockerWriter.ts
@@ -1,5 +1,6 @@
 import { FabloConfigExtended } from "../types/FabloConfigExtended";
 import { renderTemplate, getTemplatePath, getDestinationPath } from "../utils/templateUtils";
+import { shellQuote } from "../utils/shellQuote";
 import * as fs from "fs-extra";
 import * as path from "path";
 
@@ -9,7 +10,10 @@ export class FabricXDockerWriter {
   public async write(configExtended: FabloConfigExtended): Promise<void> {
     this.log("Generating Fabric-X network files...");
 
-    const data = configExtended as unknown as Record<string, unknown>;
+    const data = {
+      ...(configExtended as unknown as Record<string, unknown>),
+      shellQuote,
+    };
 
     await this.renderTemplateFile("fabric-x-docker.sh", data);
     await this.renderTemplateDirectory("fabric-x", data);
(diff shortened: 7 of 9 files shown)


MaryamArif7:
Hi @umegbewe  Id like to clarify a few things regarding this PR

1. The issue requires that Fablo be able to create at least one namespace. Should we enforce this by making namespaces a required field in the config like the network simply cannot be generated without at least one namespace ?

2. The current plan for namespace init is that it creates every namespace declared in the config, one after another. If a user has multiple namespaces defined, should we also support creating just one specifically where we create nampespace for one ( namespace init <name>) instead of only being able to create all of them at once?

3. currenlty every fabric-x-only rule (one channel, no chaincodes, TLS required) is  in validate/index.ts, not the schema. So I added namespaces as a plain optional field in the same shared schema, with the "at least one, for fabric-x" rule also enforced in validate/index.t.So here im going with  "no separate Fabric-X schema" requirement mentioned in the issue,by just applying that rule only for fabric-x.......

4.Right now we are doing something like this below :Just like this ref Make File [](https://github.com/hyperledger/fabric-x-samples/blob/main/devnet/Makefile)


```
"namespaces": [
  { "name": "mynamespace", "policy": "AND('Org1MSP.member','Org2MSP.member')" }
]

```
But the flaw with this approach is that 

we type the ploicies by hand everytime we add them even tho fablo already knows every org's MSP name from orgs[]

 what do you think if we should go with this ?
 
we can do somethig like this 
`{ "name": "mynamespace", "orgs": ["Org1", "Org2"] }`

with this we can add a helper fun ,resolvePolicy which can input the orgs with namespace names and return this 
`{ "name": "mynamespace", "policy": "AND('Org1MSP.member','Org2MSP.member')" }` it can be flow to the basefunctions

umegbewe:
@MaryamArif7 

1. Do not make namespaces globally required in the JSON schema, because classic Fabric configs should not need it. But yes, for `provider == fabric-x` validation should require at least one namespace, `fablo init fabric-x` should generate one by default

2. namespace init should create all configured namespaces by default. I’d also support namespace init <name> if it is not too much extra work, because once multiple namespaces exist, users will expect targeted creation

3. Yes, provider-specific rules belong in validate/index.ts, not as global schema requirements. The schema should define the optional top-level namespaces shape. Fabric-X validation should enforce “at least one namespace,” uniqueness, and valid org references.

4. Yes, use orgs, not hand-written policy as the normal interface. Something like

```json
"namespaces": [
  { "name": "mynamespace", "orgs": ["Org1", "Org2"] }
]
```

advanced override can be

```json
"namespaces": [
  { "name": "custom", "policy": "AND('Org1MSP.member')" }
]
```

I’d avoid allowing both orgs and policy in the same namespace for now

MaryamArif7:
implementation is completed ,only testing is left

src/extend-config/extendConfig.ts
```diff
@@ -6,14 +6,15 @@ import extendChannelsConfig from "./extendChannelsConfig";
 import extendChaincodesConfig, { checkUniqueChaincodeNames } from "./extendChaincodesConfig";
 import extendHooksConfig from "./extendHooksConfig";
 import { distinctOrdererHeads, mergeOrdererGroups } from "./mergeOrdererGroups";
-
+import extendNamespacesConfig from "./extendNamespacesConfig ";
```
umegbewe: there's a space here, likely why this PR is failing

src/extend-config/defaults.ts
```diff
@@ -34,4 +34,9 @@ export default {
       return capabilities.isV2 ? undefined : `AND (${orgs.map((o) => `'${o.mspName}.member'`).join(", ")})`;
     },
   },
+   namespace: {
+    policy(orgs: OrgConfig[]): string {
+      return `AND(${orgs.map((o) => `'${o.mspName}.member'`).join(",")})`;
+    },
+  },
```
umegbewe, line 41: I highly doubt this compiles, i don't see OrgConfig being imported

src/extend-config/extendConfig.ts
```diff
@@ -25,7 +26,7 @@ const extendConfig = (json: FabloConfigJson): FabloConfigExtended => {
   checkUniqueChaincodeNames(chaincodesJson);
   const chaincodes = extendChaincodesConfig(chaincodesJson, channels, global);
   const hooks = extendHooksConfig(hooksJson);
-
+  const namespaces = extendNamespacesConfig(namespacesJson ?? [], orgs);
```
umegbewe:
```suggestion
  const namespaces = global.provider === "fabric-x" ? extendNamespacesConfig(namespacesJson ?? [], orgs) : [];
```

src/commands/validate/index.ts
```diff
@@ -583,6 +583,41 @@ export default class Validate extends Command {
...
+    const namespaces = networkConfig.namespaces ?? [];
+    if (namespaces.length < 1) {
+      this.emit(validationErrorType.ERROR, {
+        category: validationCategories.GENERAL,
+        message: "fabric-x requires at least one entry in 'namespaces'.",
+      });
+    }
+ 
+    const namespaceNames = new Set<string>();
+    const knownOrgNames = networkConfig.orgs.map((o) => o.organization.name);
```
umegbewe: This validates namespace orgs against all declared orgs, not the orgs in the Fabric-X channel. That can generate a policy for an MSP that is not actually part of the generated Fabric-X channel config

For orgs, validate against channels[0].orgs, not only networkConfig.orgs, unless the generator is also changed to include every declared org in configtx.yaml

src/commands/validate/index.ts
```diff
@@ -583,6 +583,41 @@ export default class Validate extends Command {
...
+      namespaceNames.add(namespace.name);
+ 
+      if (namespace.orgs && namespace.policy) {
+        this.emit(validationErrorType.ERROR, {
+          category: validationCategories.GENERAL,
+          message: `Namespace '${namespace.name}' defines both 'orgs' and 'policy'. Use only one.`,
+        });
+      }
+ 
+      const unknownOrgNames = (namespace.orgs ?? []).filter((n) => !knownOrgNames.includes(n));
```
umegbewe, line 631: Same issue here

src/extend-config/extendNamespacesConfig.ts
```diff
@@ -0,0 +1,54 @@
...
+  if (policy) {
+    return policy;
+  }
+
+  if (namespaceOrgNames) {
+    const knownOrgNames = orgsTransformed.map((o) => o.name);
+    const unknownOrgNames = namespaceOrgNames.filter((n) => !knownOrgNames.includes(n));
+    if (unknownOrgNames.length > 0) {
+      throw new Error(`Namespace '${name}' references unknown org(s): ${unknownOrgNames.join(", ")}.`);
+    }
```
umegbewe, line 28: `orgs: []` is truthy and will produce `AND()`, which is not a useful policy. Add `minItems: 1` in the schema and a validation error for empty namespace.orgs.

src/types/FabloConfigExtended.ts
```diff
@@ -208,4 +211,5 @@ export interface FabloConfigExtended {
   channels: ChannelConfig[];
   chaincodes: ChaincodeConfig[];
   hooks: HooksConfig;
+  namespaces: NamespaceConfig[]
```
umegbewe:
```suggestion
  namespaces: NamespaceConfig[];
```

src/commands/init/index.ts
```diff
@@ -113,9 +113,15 @@ function getDefaultFabricXConfig(): FabloConfigJson {
...
+      {
+        name: "mynamespace",
+      },
+
+    ],
     chaincodes: [],
     hooks: {},
-  };
+  }
 }
```
umegbewe, line 124: Since the standard path is now org-based policy generation, the default Fabric-X config should show that:

```
namespaces: [
  {
    name: "mynamespace",
    orgs: ["Org1"],
  },
],
```

also clean that extra blank line and keep };

umegbewe:
This PR needs tests @MaryamArif7 

add tests for:

* `extendNamespacesConfig` derives `AND('Org1MSP.member','Org2MSP.member')` from orgs.
* policy override is preserved.
* duplicate namespace names fail.
* orgs and policy together fails.
* unknown org fails.
* empty orgs: [] fails.
* generated base-functions.sh creates all namespaces by default.
* namespace init <name> creates only the selected namespace and errors for unknown names.

Good work so far!

src/extend-config/defaults.ts
MaryamArif7: its there at line 1 "import { Capabilities, OrdererGroup, OrgConfig } from "../types/FabloConfigExtended";"
i think it was outside the diff view

MaryamArif7:
I've resolved the changes , I'll add the tests and test everything thoroughly locally ,currently its failing because snapshots are not updated and namespace init is failing

MaryamArif7:
@umegbewe 
**Why the namespace name validation was added**

While testing, a namespace named audit-ns failed at runtime. tracing it back to the actual Fabric-X source code showed that namespace IDs are restricted to lowercase letters, digits, and underscores only, no hyphens, with a 60 character limit. That rule lives in fxconfig ,so i have added those validation in valid/index file,extendedconfig and tests files.

**Local testing**

**With the default namespace**

**Namespace init**

 image

**Namespace list** 

image

**With multiple namespaces added**

**Namespace init**

This initializes every namespace declared in the config file.

image

image

**Namespace list**

This shows all the namespaces that were created.

image

**Targeted namespace init**

**./fablo.sh namespace init audit_ns**

image

image

**Validation checks** 

image

src/extend-config/extendConfig.ts
umegbewe: This is not resolved @MaryamArif7

MaryamArif7:
> This PR needs tests @MaryamArif7
> 
> add tests for:
> 
> * `extendNamespacesConfig` derives `AND('Org1MSP.member','Org2MSP.member')` from orgs.
> * policy override is preserved.
> * duplicate namespace names fail.
> * orgs and policy together fails.
> * unknown org fails.
> * empty orgs: [] fails.
> * generated base-functions.sh creates all namespaces by default.
> * namespace init  creates only the selected namespace and errors for unknown names.
> 
> Good work so far!

hi @umegbewe 
All covered in the added test files. 
extendNamespacesConfig.test.ts:
covers deriving policy from orgs, policy override, duplicate names, orgs+policy together, empty orgs, and unknown org. 

namespaceInit.test.ts
 covers the shell side: init creates all namespaces by default, targeted init creates only one, and an unknown name fails cleanly. Also updated test-08-fabric-x.sh with targeted init and unknown name checks.

e2e-network/docker/test-08-fabric-x.sh
```diff
@@ -79,6 +79,13 @@ run_fablo namespace init
 # post-init query
 expectCommand "(cd \"$TEST_TMP\" && \"$FABLO_HOME/fablo.sh\" namespace list)" "mynamespace"
 
+# targeted init: creating an already-existing namespace by name should still succeed (idempotent)
+run_fablo namespace init mynamespace
+expectCommand "(cd \"$TEST_TMP\" && \"$FABLO_HOME/fablo.sh\" namespace list | grep -c \"mynamespace\")" "1"
+
+# targeted init with an unknown name should fail, not hang or silently succeed
+expectNotCommand "(cd \"$TEST_TMP\" && \"$FABLO_HOME/fablo.sh\" namespace init does-not-exist || echo \"EXIT:\$?\")" "EXIT:0"
```
umegbewe: When the command succeeds, `|| echo ...` outputs nothing, so this assertion still  anyways

Positively assert both the error message and a non-zero exit result, also maybe use a syntactically valid but unknown namespace such as does_not_exist:

```sh
expectCommand "(cd \"$TEST_TMP\" && \"$FABLO_HOME/fablo.sh\" namespace init does_not_exist 2>&1 || echo \"EXIT:\$?\")" "Namespace 'does_not_exist' not found"
expectCommand "(cd \"$TEST_TMP\" && \"$FABLO_HOME/fablo.sh\" namespace init does_not_exist 2>&1 || echo \"EXIT:\$?\")" "EXIT:"
```

src/setup-docker/namespaceInit.test.ts
```diff
@@ -0,0 +1,57 @@
...
+import { execSync } from "child_process";
+import { shellQuote } from "../utils/shellQuote";
+
+describe("fabric-x base-functions.sh namespaceInit", () => {
+  const templatePath = path.join(__dirname, "templates/fabric-x/scripts/base-functions.sh");
+  const template = fs.readFileSync(templatePath, "utf-8");
+
+  const namespaces = [
+    { name: "mynamespace", policy: "AND('Org1MSP.member')" },
+    { name: "audit-ns", policy: "OutOf(1, 'Org1MSP.member')" },
```
umegbewe: The schema says hypens on namespaces are rejected though 🤔

docs/schema.json
```diff
@@ -624,6 +624,54 @@
...
+      "properties": {
+        "name": {
+          "$id": "#/properties/namespaces/items/properties/name",
+          "type": "string",
+          "title": "Name",
+          "pattern": "^[a-z0-9_]+$",
+          "maxLength": 60,
+          "description": "Must match ^[a-z0-9_]+$ (lowercase letters, digits, underscores only - no hyphens), max 60 characters. This is enforced by the Fabric-X fxconfig namespace ID validation, not just Fablo's own convention."
+        },
+        "policy": {
```
umegbewe, line 657: add minLength = 1 to policy

src/setup-docker/namespaceInit.test.ts
MaryamArif7: i have fixed it now ,that was a leftover example from before the namespace name validation was added. This test only checks the shell script logic, not validation, so it never caught it. Fixed to audit_ns.

umegbewe (approved):

PR #814, opened 2026-08-27
feat: implement issues from maintainer comments

Efeguclu1:
- add a maintainer-only issue implementation command
  - orchestrate Claude Code through Jaiph
  - run build and unit verification with recovery attempts
  - open a signed-off pull request for successful or partial implementations
  - prevent agent changes to workflows, Jaiph definitions, credentials, and
  Git history
  - document maintainer usage and add Jaiph workflow CI

  ## Trigger

  A maintainer comments one of these commands as the first line of an open
  issue:

      @fablo-bot implement

  or:

      /implement

  Additional lines are passed to the agent as implementation guidance.

PR #829, opened 2026-09-10
fix: address  feedback on the implement-issue bot

Efeguclu1:
- The implementation runs on Opus instead of Sonnet.
- The agent is pointed at CONTRIBUTING.md, and the snapshot instruction is fixed. It named \`npm run test:e2e\` and discouraged updating snapshots, which is why #825 hand-wrote three of them and CI failed on \`test-main\`. Snapshots are generator output, so the workflow now regenerates them with \`npm run test:e2e-update\` for any change that is not docs-only, and \`npm run lint\` runs alongside the existing build and unit checks.
- Steering typed on the same line as \`/implement\` is no longer discarded, \`/implementation...\` no longer triggers a run, CRLF comment bodies are handled, and a stale \`ai/issue-N\` branch left by a closed pull request is cleaned up before a re-run.
- The agent may now change \`.github/\` and \`.jaiph/\`, so an issue about the CI or the Jaiph programs can be implemented and a new \`e2e-network/docker\` script can come with its own job. Any such change opens the pull request as a draft with a warning listing the affected files, since they run with repository credentials. The policy is read from the base commit, so a check cannot be edited by the run it governs.
- The CONTRIBUTING.md section added in #814 is reduced to the command and the fact that it opens a pull request; the operational detail moves into the workflow file.


.github/workflows/implement-issue.yml
```diff
@@ -109,12 +129,52 @@ jobs:
...
       - name: Enforce no destructive changes
         id: safety
         if: always() && (steps.implement.outcome == 'success' || steps.implement.outcome == 'failure')
-        run: .jaiph/verification.jh --unsafe no_destructive_changes --yes -- "${{ github.sha }}"
+        run: |
+          set -euo pipefail
+          policy="${RUNNER_TEMP}/verification.jh"
+          git show "${{ github.sha }}:.jaiph/verification.jh" > "$policy"
+          chmod +x "$policy"
+          "$policy" --unsafe no_destructive_changes --yes -- "${{ github.sha }}"
```
dzikowski: why was it changed?

dzikowski (approved):

.github/scripts/implement-issue-gate.sh
dzikowski, line 23: check if TRIGGER_ACTOR is in AI_ALLOWED_ACTORS (comma-separated list of actors)

PR #831, opened 2026-09-14
feat(fabric-x): make Fabric-X image versions configurable (#820)

### LINKED ISSUE #820: feat (fabric-x):  Make Fabric-X image versions configurable
Fabric-X uses a different set of container images from classic Fabric. Research which generated files currently hardcode Fabric-X image names or tags and decide how those values should be configured in Fablo.

The default image versions should remain pinned and working, but users should have a clear supported way to override Fabric-X image versions without changing classic Fabric behavior.

HarK-github:
Fixes #820

Makes Fabric-X container image versions configurable via `global.fabricImages` in `fablo-config.json` while maintaining working pinned defaults.

### Changes
- **Configuration & Types**: Added `committer` and `postgres` to `FabricImagesJson` and resolved `ordererImage`, `committerImage`, `toolsImage`, and `postgresImage` in `extendGlobal.ts`.
- **Templates**: Replaced hardcoded Fabric-X image tags in `docker-compose.yaml` and `scripts/base-functions.sh` with template variables from `global`.
- **Validation & Schema**: Updated `docs/schema.json` and added warnings for mismatched provider/image overrides.
- **Tests**: Unit tests in `extendGlobal.test.ts` and generator test in `e2e/fabloCommands.test.ts`.


src/types/FabloConfigExtended.ts
```diff
@@ -10,14 +10,16 @@ export interface FabricVersions {
...
-  baseosImage: string;
-  javaenvImage: string;
-  nodeenvImage: string;
+  ccenvImage?: string;
+  baseosImage?: string;
+  javaenvImage?: string;
+  nodeenvImage?: string;
+  committerImage?: string;
+  postgresImage?: string;
 }
```
umegbewe, line 23: Don't weaken this, revert the optional and create a new interface

```ts
export interface FabricXImages {
  ordererImage: string;
  toolsImage: string;
  committerImage: string;
  postgresImage: string;
}
```

umegbewe:
Looks good add tests for proving

- Fabric-X warns when a classic-only image is configured.
- Classic Fabric warns when committer is configured.
- Shared orderer and tools settings do not produce warnings.

